PR #7, opened 2019-04-01
Changed .hgignore to .gitignore. A number of fixes for the OpenMDAO wrapper.

anilyil:
Previous pull req got screwed up since adflow went public. This should have the proper changes from the public fork of adflow.


anilyil:
Also added capability for the ank solver to handle turbulence models with multiple working variables

PR #12, opened 2019-05-23
Zipper fix and some other modifications

anilyil:
Fixed the call order on the reverse zipper integration.
Added area averaged total and static pressure integration.
Bug fix with area integration with overset meshes.
Bunch of additions to .gitignore.
Minor fix in the ANK solver physicality check.
Modified how the restarts are handled with the OpenMDAO states component.
Minor fix for edge case when we are using actuator regions and the previous iteration was partially converged.


src/solver/solvers.F90
```diff
@@ -1014,6 +1014,9 @@ subroutine solveState
     ! call, "Iter" and "Iter Total" will both display 0.
     approxTotalIts = 0
 
+    ! we need to re-set the orders converged to 16 as it might have been modified in the previous iteration
+    ordersConverged = 16.0_realType
```
gkenway: Why is this necessary exactly? The ordersConverged gets set unconditionally below anyway.

anilyil:
Also added a fix for an edge case, where ANK segfaults with Euler and turbKSP turned on. If we are not doing RANS, ank setup will not try to create the objects for the turbKSP

eirikurj:
Is something here still pending? @anilyil do we need additional reviews, new regression tests for these cost functions?

anilyil:
@eirikurj @camader Can you verify my changes? I added the area averaged ps and ptot costfunctions to regression tests 16 to 18. I initially used the results from these tests to verify the implementation. I have re-trained the reg tests 16-18, please run them yourselves

camader:
@anilyil did you see/address gaetan's comment?

anilyil:
Yeah, you can see my reply to him after his comment. I tried to explain the logic behind having that there. It only affects a very small subset of cases where we have an actuator region, and restarting from a partially converged solution.

PR #26, opened 2020-03-13
Adding in the new tests with testflo

nbons:
Everything passes except test18 which for some reason only works if you don't use the restart file (solve=True).


nwu63:
Did you verify that the new reference values match those from before? If so it would be great to have this workflow documented somewhere for others who are also migrating to testflo.

joanibal:
@nbons  the reference values do match right? Except for test 18 which does match if you don't use the given restart file (just like the normal regtest)

@nwu63  It would be great to have the workflow documented, but having this example is better than having nothing in the meantime. This pull request shouldn't  depend on the documentation of our testing workflow, right?

nwu63:
I'm fine with merging this as long as the reference values have been checked by someone else that they match. That's why I asked for the command/process to check it. We're only going to do this once per repo and I want to make absolutely sure that the integrity of the tests remain, since that's all we've got to verify our code.

anilyil:
@nbons can you change the target branch to mdolab/testflo from mdolab/master? 
If you can't, I will close this PR and you can create a new PR to the testflo branch.

This is a great effort to move the testing suite to testflo; however, it is not complete at all. @nwu63 and I decided that putting this code in a separate branch until all the testing stuff is ready to be done this way would be a good idea to get it off of your plate and keep developing it.

I have a number of issues with the current implementation, and I will create a github issue to have a conversation going on these developments.

joanibal:
@anilyil can you add your issues soon. I'm working on this some.

anilyil:
See issue #28

PR #27, opened 2020-03-16
Maintenance code updates

camader:
## Purpose
This pull request adds functionality for overset cut callbacks and fixes a number of small bugs.

## Type of change
What types of change is it?

## Testing
Should function with existing testing

## Checklist

- [ x] I have run unit and regression tests which pass locally with my changes
- [n/a] I have added new tests that prove my fix is effective or that my feature works
- [n/a ] I have added necessary documentation


python/pyADflow.py
```diff
@@ -1682,6 +1692,146 @@ def Func(Xn, CLstar):
...
+        nIter : int
+            Maximum number of iterations.
+        Jac0 : nxn numpy array
+            Initial guess for the func-dv jacobian. Usually obtained
+            from a previous analysis and saves n function
+            evaluations to produce the intial jacobian.
+        """
+
+        self.setAeroProblem(aeroProblem)
+        apDVList = ['alpha','altitude']
```
anilyil: is the apDVList limited to these variables? What about mach, etc.?

python/pyADflow.py
```diff
@@ -1682,6 +1692,146 @@ def Func(Xn, CLstar):
...
+                print ("New J:", J)
+
+            # Shuffle the Fn and Xn backwards
+            Fn = Fnp1.copy()
+            Xn = Xnp1.copy()
+
+            # Check for convergence
+            if numpy.linalg.norm(Fn) < tol:
+                if self.comm.rank == 0:
+                    print ('Converged!', jj, Fn, Xn)
```
anilyil: maybe also print what jj, Fn, and Xn are? Also, Fn is just the error for each function, maybe the actual value would also be useful?

python/pyADflow.py
```diff
@@ -1682,6 +1692,146 @@ def Func(Xn, CLstar):
...
+        """
+        Solve the an arbitrary set of function-dv sets using a Broyden method.
+
+        Parameters
+        ----------
+        AeroProblem : AeroProblem instance
+            The aerodynamic problem to be solved
+        funcDict : dict
+            Dictionary of function DV pairs to solve...add dv index {'func':{'dv':str, 'dvIdx':idx,'target':val,'initVal':val,'initStep':val}}
+        tol : float
```
anilyil: can you be a bit more clear at what this tolerance is? The current docstring is whats left from another function. We actually do the L2 norm of the error vector. So individual functions are not guaranteed to be solved to the tolerance. We may want to include an option to check for each function individually. Maybe scaling is bad on a function that does not converge very tightly, and a residual on the order of "tol" is not good enough, yet the norm is less than the norm due to other functions etc.

python/pyADflow.py
```diff
@@ -1682,6 +1692,146 @@ def Func(Xn, CLstar):
...
+    def solveTargetFuncs(self, aeroProblem, funcDict, tol=1e-4, nIter=10, Jac0=None):
+        """
+        Solve the an arbitrary set of function-dv sets using a Broyden method.
+
+        Parameters
+        ----------
+        AeroProblem : AeroProblem instance
+            The aerodynamic problem to be solved
+        funcDict : dict
+            Dictionary of function DV pairs to solve...add dv index {'func':{'dv':str, 'dvIdx':idx,'target':val,'initVal':val,'initStep':val}}
```
anilyil: can you explain the keys individually? the dvIdx in particular can be more clear. dv can be the full AP dv name, or just "alpha" etc., and in that case, we append the AP name. Maybe add a TODO comment to add a scaling option for a func as well, since they may not be scaled evenly and we currently check the L2 norm.

python/reg_tests/commonUtils.py
```diff
@@ -367,7 +367,7 @@ def standardTest(CFDSolver, ap, solve):
     reg_par_write_norm(resDot, 1e-10, 1e-10)
 
     parPrint('dFuncs/dw * wDot')
-    reg_root_write_dict(funcsDot, 1e-10, 1e-10)
+    reg_root_write_dict(funcsDot, 1e-9, 1e-9)
```
anilyil: does this have any implications? Will the tests that we are currently moving to other testing formats also contain this change?

python/pyADflow.py
camader: I can modify this to pull the possible DVs from the aeroproblem

python/pyADflow.py
camader: We never have access to  the actual functions in this spot, so not super easy to do

python/pyADflow.py
camader: I have updated the comment. I don't think this function is used in a way that makes individual scaling or error checking meaningful

python/pyADflow.py
camader: extended comment

python/reg_tests/commonUtils.py
camader: this was a mistake. Should not have been changed

PR #29, opened 2020-04-07
Python 3 updates

nwu63:


nwu63:
That is a good question, I'm not sure why that got added, probably because I did this a long time ago and didn't properly merge. I've done that now and hopefully the tests pass.

anilyil:
Is there a way to not hard-code the python-configure stuff in config files? e.g.: https://github.com/mdolab/adflow/pull/29/files#diff-b7b1101023bf8d7d75af329bdd38f9a1R59 This will probably break installation with python2 by default correct?

nwu63:
Correct. Part of the reason for this was that _by default_ people _should_ be using python 3, and will require them to change one line in the config file to use python 2. I didn't want to add more configuration files since that would just bloat things and cause code duplication, and we already do this with things like CGNS, having a default option and asking people to change it if they use a non-default setup.

We could also use templating via jinja to generate these configuration files on the fly similar to a configure step, or we could pick out some sort of environment variable that we ask everyone to set, but both have shortcomings too. Thoughts?

anilyil:
If you think people should switch to python 3, that is fair. But, there are a lot of people who use py2 for other stuff and cannot switch. I agree with you on not creating more config files.

Changing the config files like this will have a weird effect. For people who already have compiled ADflow, they will pull this, and because they copied the config file way back, it will still work for the most part. Then, when they try to install it from scratch, it will break and they wont understand why.

Again, I agree with your rationale. If you make this change, you need to loudly communicate this to people so that they do not need to waste time trying to figure out what is broken. This is my main concern.

nwu63:
I think it's pretty reasonable to default to python 3, and we will probably drop python 2 some time later this year anyways. 

Fair, we can talk about ways to communicate well. In the longer term I think the templating approach may be better than having a bunch of configuration files lying around (with minimal documentation on which one they should use). But we would have to do some work to develop that (and make it consistent across our codes).

nwu63:
@eirikurj added some notes in the installation docs [here](http://mdolab.engin.umich.edu/docs/installInstructions/installFromScratch.html#standard-mdo-lab-build-procedure), which should address your concerns @anilyil.

anilyil:
Okay, that looks good to me

Issue #28, opened 2020-03-29
Migration to Testflo for regression testing


anilyil:
## Type of issue

- Other: Code modifications required to change the underlying testing suite to testflo.

## Description

This issue will contain the discussions regarding the migration of adflow testing to testflo. The modifications in PR #26 to migrate the testing suite to testflo is missing a number of important features. The development related to this will be pulled into the testflo branch of adflow, and the changes will be merged to the master branch once all expected functionality is provided.

### Current behavior

Here is a list of issues (in no particular order) that I think should be addressed. This is not an exhaustive list and we will have the discussions related to these items in this issue page.

1. Individual tests do not provide feedback as they finish, which makes it difficult to determine the current progress.
2. Tests don't run in order, e.g. test 18 runs before test 1, etc.
3. No complex tests.
4. No output from tests. The output from the test runs should go into a file for each test for easier debugging.
5. No documentation on obtaining the input files required for tests. Existing documentation for input files is in a readme file in a directory that is a few levels down the main directory. There are no external links (that I could see) to this readme. The directory name is changed from inputFiles to input_files
6. It's not clear how many processors is used for each test. This may be related to testflo, but either way, this should be clearly documented.
7. If multiple tests run at the same time, it should be guaranteed that tests do not try to over-write to same output files.
8. The tests generate a lot of intermediate files that are not cleaned up and appear in the git status.
9. No documentation in general for simple testflo options.
10. If a value does not match in the reg tests, the output should also state what the value means (dot product test of this routine etc.).
11. Old test file and folder structure are still there.

### Expected behavior

Here is a list of expected behavior for all the issues I raised above (with the same numbering):

1. Individual tests should provide feedback as they finish, regardless of them failing or succeeding. 
2. Tests should run in order, e.g. test 1, then 2, then 3, up to 18. Although this is not strictly required, it will become tricky to figure out when a particular test ran if this order is not followed. If tests run in order, I know that test 3 did not run yet since test 2 is still running. If this order is not preserved, I would need to go through the full output to see if that particular test ran already. This will become a larger issue as we add more tests.
3. The suite should also contain complex tests. 
4. Each test should write its output to an output file for easier debugging. These files can be put in a folder, which should be included in the gitignore list.
5. Documentation needs to be greatly improved. The readme file in the root adflow directory should point to the testing documentation directly. This documentation should include directions on how to obtain the input files. The documentation should also include various testflo options that can be used to run the tests, and possibly contain a link to the testflo documentation itself.
[...]

PR #30, opened 2020-04-09
iteration on testing setup with an example 

joanibal:
## Purpose
This PR gives an example of modifications to the tests in an effort to resolve 
[this issue](https://github.com/mdolab/adflow/issues/28). 
Please examine reg_tests/test_euler_tutorial_wing.py and provide feedback on the modifications to the testing structure. 

This PR also fixes a bug in the reverse mode routine that caused the example test to fail 

## Type of change
1) The test handler uses a dictionary to store the reference values
2) Regression tests are combined into the a single based on the case
3) Each regression test is renamed and split up into a set of smaller tests 
   to emphasize clarity

- Breaking change (non-backwards-compatible fix or feature)
- Code style update (formatting, renaming)

## Testing
Pull the testflo branch of baseclasses on my fork. 
run then test_euler_tutorial_wing.py

## Checklist


src/adjoint/masterRoutines.F90
```diff
@@ -607,6 +609,56 @@ subroutine master_b(wbar, xbar, extraBar, forcesBar, dwBar, nState, famLists, &
...
+            ! Make sure to call the turb BC's first incase we need to
+            ! correct for K
+            if (equations == RANSequations) then
+               call BCTurbTreatment
+               call applyAllTurbBCthisblock(.True.)
+            end if
+            call applyAllBC_block(.True.)
+
+
+            call timeStep_block(.True.)
```
joanibal, line 637: @camader this is the bug fix (line 637) I referenced in the meeting.
In addition to resetting the boundary conditions the spectral radius needs to be calculated since it isn't calculated in the residual calculation. 

But I'm a bit confused why the spectral radius isn't set by calling the getResiduals() routine on the python level

joanibal: It looks like it is an issue with using the blockettes

anilyil: can you explain the bug here as well?

nwu63:
I personally think this should be split into two PRs, one for testing and one for bug-fix. The PR for bug fix should be made to the master branch and not the testflo branch.

joanibal:
I think that makes sense. 
I'd mostly just like to get feedback on the the testing script.

src/adjoint/masterRoutines.F90
joanibal: After you instantiate an ADflow instance with a restart file.
If you call 
`CFDSolver.getResiduals(ap)
CFDSolver.computeJacobianVectorProductBwd(....)`
you get the wrong answer

But if you call 
`CFDSolver.getResiduals(ap)
CFDSolver.computeJacobianVectorProductFwd(....)`
CFDSolver.computeJacobianVectorProductBwd(....)`
you get the right answer

If you trace the error back you see that this is because the spectral radii (radI, radJ, radK) aren't set by getResiduals() with the default options

I just learned that if you set the option `useblockettes' to False there is no issue. So it seems like the spectral radii are being zeroed before use in master_b when using blockettes.

Per neil's suggestion I'm going to make a separate issue to discuss this further

tests/README.md
```diff
@@ -2,9 +2,25 @@
...
+- To train all regression tests in specific testfile either works: `cd reg_tests; testflo test_DVGeometry.py -m "train_*"` or `testflo reg_tests/test_DVGeometry.py -m "train_*"` 
+- To train specific regression test in specific testfile either works: `cd reg_tests; testflo test_DVGeometry.py -m "test_1` or `testflo reg_tests/test_DVGeometry.py -m "train_1"`
+
+- To test all regression tests: `testflo reg_tests`
+- To test all regression tests with specific function signature: `testflo reg_tests -m "test_*"`
+- To test all regression tests in specific testfile either works: `cd reg_tests; testflo test_DVGeometry.py -m "test_*"` or `testflo reg_tests/test_DVGeometry.py -m "test_*"` 
+- To test specific regression test in a specific testfile either works: `cd reg_tests; testflo test_DVGeometry.py -m "test_1` or `testflo reg_tests/test_DVGeometry.py -m "test_1"`
+
+- To enable screen output: use `-s` argument
+- To change number of cores to 1 for running tests: use `-n 1`
```
nwu63, line 17: It would be useful to provide a single command which would run all ADflow tests, and return the result to screen.

tests/README.md
```diff
@@ -2,9 +2,25 @@
...
+
+## Other ways to run
+The tests are also compatiable with general ```unittesting``` python framework and can be call individually using, 
+
+```python unit_tests/test_basics.py```
```
nwu63, line 22: Will this work with MPI, or are you simply running the same test but with one proc?

tests/reg_tests/commonUtils.py
```diff
@@ -248,6 +253,282 @@ def adjoint_test(handler, CFDSolver, ap):
     handler.root_print('Eval Functions Sens:')
     handler.root_add_dict(funcsSens, 1e-10, 1e-10)
 
+def assert_adjoint_sens_allclose(handler, CFDSolver):
```
nwu63, line 256: Would it be possible to add atol and rtol as optional arguments to each of these functions? I think that would be useful to have the option instead of hard-coding everything as 1e-10.

tests/reg_tests/test_euler_tutorial_wing.py
```diff
@@ -0,0 +1,193 @@
...
+class TestSolution(unittest.TestCase):
+    '''
+    Test 9: MDO tutorial -- Euler -- Solution Test
+    '''
+    N_PROCS = 4
+
+    def setUp(self):
+
+        self.ref_file = os.path.join(baseDir, 'ref/ref9.py')
+        ref = run_path(self.ref_file)['ref']
```
nwu63, line 26: Instead of using `run_path`, I think you can just import the file, using something like

`from ref.ref9 import ref`

tests/reg_tests/test_euler_tutorial_wing.py
```diff
@@ -0,0 +1,193 @@
...
+sys.path.append(os.path.join(baseDir,'../../'))
+from python.pyADflow import ADFLOW
+import copy
+from baseclasses import AeroProblem
+# from reg_tests.commonUtils import standard_test, adflowDefOpts, defaultFuncList
+from commonUtils import *    
+from runpy import run_path
+
+
+class TestSolution(unittest.TestCase):
```
nwu63, line 17: It would be useful to have a more descriptive name here.

nwu63:
I just thought of something else. What do you think about having reference files in JSON format? It's a readable plain text format that is supported natively in Python and across many other languages.

tests/README.md
joanibal: you can run all the regression tests with `testflo reg_tests -s`.
Is that what you mean?

tests/README.md
joanibal: It wouldn't work with MPI, it uses just one proc

tests/reg_tests/test_euler_tutorial_wing.py
joanibal: I was doing that originally, but i wanted to make sure training wrote to the same file so i thought it was better to use the file path when importing the reference

tests/reg_tests/test_euler_tutorial_wing.py
joanibal: How about TestSolve? 
That test checks to make sure adflow can still solve the case to the same tolerance and that the solved state and functions are consistent with the reference.
Ideally we'd added more detail to the test description too.

joanibal:
I thought about using JSON, but decided not to since i was less familiar with it. 
The read and write of the reference file are separated from the handler, so it shouldn't be an issue if we wanted to do that. 

I gave that some thought too. I think the tolerances shouldn't be stored in the reference data since it would make the reference file more complicated. In the old framework the tolerances originated in the tests script but were propagated to the reference and output because it was needed for comparison. I don't think having them in the output adds much value in and of itself, but it can always be a feature we add later.

tests/README.md
nwu63: Right, I don't think it would make sense to have that there then, since it's not going to run the test with the correct configuration. Since we can call individual tests with testflo we don't need to run it with the unittest framework directly.

tests/reg_tests/test_euler_tutorial_wing.py
nwu63: Right that's a good point, okay.

tests/reg_tests/test_euler_tutorial_wing.py
nwu63: Yes agreed. I saw that there's a standard_test, is that basically the same thing as what's done here?

nwu63:
Hmm okay, I do think there are benefits to using JSON since it's a standard data format, but I don't feel super strongly about it. I just think if we want to switch now would be the time to do that, since it's likely once we set things in place they would not be changed for a very long time.

tests/README.md
nwu63: I haven't tried that to see what it would output, but basically a command that replaces the `python run_reg_tests.py` command we had before, such that ADflow users can verify their installation. We can have additional commands for developers and others to debug things, but it would be useful to recommend a single command to use just to verify everything.

tests/reg_tests/test_euler_tutorial_wing.py
joanibal: This specific test would be analogous to solution_test in the old framework.
TestFuntionals performs the same tests as standard_test

tests/README.md
joanibal: I found using unittesting directly to be very valuable. Specifically when I wanted to call one sub test for debugging (test_tutorial_wing.TestFunctionals.test_jac_vec_prod_bwd for example).
Testflo might be able to do this but I haven't figured it out. 

would you support rewording to something like ...

> The tests can also be called without multi-core execution using unittesting directly with
```python unit_tests/test_basics.py```
or any other way to call the unittesting package. 
This method of running the tests should only be used for debugging the tests since it is not compatible with MPI and will not catch issues with parallel execution.

File: tests/README.md (hunk 2 above)
Comment (joanibal) [reply tests/README.md:17]: I think the best option would be to just make a new `run_reg_tests.py` even if it just makes a call to the shell.That would make it clear for those just looking to run all the tests.

People who want more options can use the README.md to figure out what to pass to testflo.

**Comment by joanibal** [comment]
I'll give JSON a try and let you know if i have any trouble

File: tests/README.md (hunk 3 above)
Comment (nwu63) [reply tests/README.md:22]: It is possible to run individual tests. For example in pyoptsparse, you could run something like `testflo test/test_hs015.py:TestHS15.test_snopt` to run that specific test. We should definitely explain this syntax in our README though.

Would that be sufficient to replace the unittest stuff?

Comment (joanibal) [reply tests/README.md:22]: Yeah that would work. I'll remove any reference to unittesing

**Comment by joanibal** [comment]
@nwu63  Please review the changes made based on your feedback.

File: tests/run_all_tests.py [hunk 7]
```diff
@@ -0,0 +1,7 @@
+import os 
+
+print('=========================== running unit tests ========================')
+os.system('testflo unit_tests -v')
```
nwu63, line 4: I think the default flags we want to use is probably something like `-v -n 4`, since otherwise testflo treats logical cores as physical cores. This would be consistent with the previous `mpirun -np 4` command in `run_reg_tests.py`.

tests/reg_tests/test_euler_tutorial_wing.py
```diff
@@ -0,0 +1,195 @@
...
+from baseclasses import BaseRegTest
+import os
+import sys
+baseDir = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.join(baseDir,'../../'))
+from python.pyADflow import ADFLOW
+import copy
+from baseclasses import AeroProblem
+# from reg_tests.commonUtils import standard_test, adflowDefOpts, defaultFuncList
+from commonUtils import *
```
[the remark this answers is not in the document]
joanibal: agreed

tests/reg_tests/test_euler_tutorial_wing.py
```diff
@@ -0,0 +1,195 @@
...
+from runpy import run_path
+
+
+class TestSolve(unittest.TestCase):
+    '''
+    Test 9: MDO tutorial -- Euler -- Solution Test
+    '''
+    N_PROCS = 4
+
+    def setUp(self):
```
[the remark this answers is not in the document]
joanibal: You could put the solve into the setUpClass, but I think it should stay as a test so that if it fails it's clear it is a problem with the code being tested not the tests.

We could remove any files that were created by the tests in the `tearDown`. What do you think about that?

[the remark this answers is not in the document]
nwu63: Oh sorry I was confused -- I thought the solve was already in the `setUp` routine. Since `setUp` is being called every time it doesn't really change anything, and you're right that this is more explicit. Given how quick these are I think it's fine as it is.

I'm not sure that makes sense, as the `tearDown` routine is called regardless of whether the test passed or not, and it's possible that the output files are useful in debugging. I was actually thinking more in terms of destroying objects and releasing memory.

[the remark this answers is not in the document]
joanibal: I'm pretty sure the __del__ method of the ADFLOW is being called after each test since I'm encountering an error there on myfork of adflow... 

So, I don't think we need to destroy anything, although I'm not 100% sure.

tests/reg_tests/test_euler_tutorial_wing.py
```diff
@@ -0,0 +1,195 @@
+from __future__ import print_function
+import unittest
+import numpy
+from baseclasses import BaseRegTest
+import os
+import sys
+baseDir = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.join(baseDir,'../../'))
```
nwu63, line 8: One thing I just noticed since I'm working on pyHyp tests, is that this append should probably be an insertion, i.e. `sys.path.insert(0,...)`. The order in `sys.path` is important, and if  the parent folder is last on the path then another set of codes could be found first, meaning you are not testing the codes you think you are testing.

Issue #2, opened 2019-03-09
Remove meshing instructions from docs


nbons:
These instructions seem out of place in the adflow docs and they are repeated in greater detail in the MACH tutorial.

joanibal:
@nbons Did this get done?

eirikurj:
This is not done. The page that he is referring to is here http://mdolab.engin.umich.edu/docs/packages/adflow/doc/meshing.html

I suggest removing it since we have a superior tutorial here http://mdolab.engin.umich.edu/docs/packages/mach_aero_tutorials/doc/aero_icem.html.

PR #33, opened 2020-04-15
Removing old docs and addressing build warnings.

eirikurj:
## Purpose
Remove docs duplication. This PR addresses issue #2 

## Type of change
- Documentation update

## Testing
Built locally

## Checklist


joanibal:
This is to address issue #2, Correct?

eirikurj:
Yes, I did mention the issue in the PR "purpose", so it should show up and be tracked with the issue as well.

joanibal:
lol, totally missed that.
I'll added at as linked issue too

Everything looks good to me

anilyil:
Instead of removing the whole section, can't we just link to the "better" documentation? Maybe, we can also include the doi numbers for the overset paper etc.?

eirikurj:
This particular section is a tutorial on how to use ICEM to generate CGNS meshes, which is unrelated to ADflow (with the exception that ADflow uses CGNS meshes).
To address your comment on adding a link, I improved the tutorial sections, by adding a link to the MACH aero tutorial.
Adding a DOI to the overset papers is also a good idea since it describes the implementation, but its not a tutorial. However, I dont think that it is relevant for the particular issue that his PR is addressing. Perhaps create a new issue which describes the improvements?

anilyil:
This PR looks good to me. I agree on adding the doi stuff for the other relevant publications also. Lets discuss the format for adding all relevant doi numbers and we can work on adding them to the adflow docs.

Issue #36, opened 2020-04-17
Spectral radii is out of date in some scenarios


anilyil:
## Type of issue

Bug

## Description

The cache-optimized residual routine (blockette residual) does not update the spectral radii arrays in the main memory. This routine compute these intermediate variables using the cached memory but does not copy the computed spectral radii values back to main memory. This later on causes issues because the reverse mode AD routines rely on the spectral radii values being up to date. 

### Steps to reproduce issue

There are multiple ways to run into this issue. I will lay out two scenarios. The issue only comes up if 'useblockettes' option is set to True, which is the current default.

#### Scenario 1: 
Analysis + adjoint
1. A regular adflow analysis (call the pyadflow object) with 'useblockettes': True, which is the default option. This way, the spectral radii (radi, radj, radk arrays) are not updated as the solver converges.
2. A call to evalFunctionsSens after the analysis will set up the adjoint solver. During this step, if the user did not specify 'adpc':True, which is not the default option, the spectral radii arrays will be wrong. Any reverse mode AD call after this point (until the fwd mode AD is run) will be wrong.
3. This will cause inaccurate sensitivities.

Going back to step 2, if the users have 'adpc':True, the code runs fwd AD routines for the preconditioner computations before any reverse AD call, which update the mentioned arrays and this avoids the problems. 

Similarly, having 'nkadpc':True with the NK solver has a similar effect. However, with this case, the arrays will still be outdated because they will be the values calculated at the last NK preconditioner computation.

If an AD preconditioner is never done with the NK or the adjoint solvers, the spectral radii arrays will have the values computed right after flow initialization and before solution starts (https://github.com/mdolab/adflow/blob/master/src/solver/solvers.F90#L1025), and will be wrong. 

#### Scenario 2: Standalone reverse AD call:

This scenario is explained in issue #32 

### Current behavior

Spectral radii is not updated in many cases.

### Expected behavior

Before any reverse mode AD call, this intermediate variable should have the correct value computed with the updated state.

### Code version (if relevant)

Independent of any code version

anilyil:
Also, this issue was not detected by the previous regression tests because they perform forward mode checks before reverse mode checks, and by the time reverse mode checks are run, the spectral radii have been updated by the fwd mode checks.

Issue #32, opened 2020-04-09
Spectral Radii zeroed after getResiduals() with blockettes


joanibal:
## Type of issue
Bug 

## Description
After you instantiate an ADflow instance with a restart file.
If you call

CFDSolver.getResiduals(ap)
CFDSolver.computeJacobianVectorProductBwd(....)

you get the wrong answer

But if you call

CFDSolver.getResiduals(ap)
CFDSolver.computeJacobianVectorProductFwd(....)
CFDSolver.computeJacobianVectorProductBwd(....)

you get the right answer

If you trace the error back you see that this is because the spectral radii (radI, radJ, radK) aren't set by getResiduals() with the default options

If you set the option `useblockettes' to False there is no issue. So it seems like the spectral radii are being zeroed before use in master_b when using blockettes.

if you add an extra call to timeStep_block at the beginning of master_b()
it fixes the issue. So i'm very certain it is because of the radii

### Expected behavior
Both should give the same answer

### Code version (if relevant)
**Python version:** 2.7
**External dependencies:** 
Petsc 3.11
OpenMPI 3.1.4

**Internal packages:** …
latest of everything

joanibal:
@anilyil

anilyil:
The ambiguity is caused by the "getResiduals" call returning only the residuals, and not guaranteeing any intermediate variable being up to date. getResiduals does work correctly, it should return the correct residuals. However, if you run it with the blockette version of the residuals, as you saw, it will not update some of the intermediate variables. 

This issue does not come up if you actually call the CFD solver. We did run into this while coding the blockette stuff, and we fixed this particular problem by adding an extra call to the timestep computation in the solvestate subroutine: https://github.com/mdolab/adflow/blob/master/src/solver/solvers.F90#L1025 Having this extra call makes sure that spectral radii are explicitly computed and saved, regardless of the blockette flag.

My proposed solution: Seems like the getRes subroutine in NKSolvers (https://github.com/mdolab/adflow/blob/master/src/NKSolver/NKSolvers.F90#L1374) is only called by the getResiduals method on the python level. We can modify this subroutine to perform an extra call to timestep if blockette flag is enabled. This way, spectral radii will be up to date. Another option is to only call the non-blockette residual in this getRes subroutine.

There are multiple ways to fix this issue. I can make this modification, but first, let me know what you think about this.

joanibal:
Why are the spectral radii not saved during getRes when using blockettes? 
[...]

PR #37, opened 2020-04-17
Copying spectral radii back to main memory after blockette routines …

anilyil:
... if updateDt is true. Without this change, reverse mode AD routines use outdated spectral radii values, which results in inaccurate sensitivities.

## Purpose

This PR addresses issues #32 and #36. 

With this change, the spectral radii values are copied back to main memory from cached memory after a blockette residual computation that also updates the time step. This copy is done inside this if check because we do not need to copy these values for matrix-free matrix-vector products; however, we want to copy them after every ANK and NK step.

The extra copy comes with additional cost due to the increase memory access. This extra cost will only be present with blockette calls that also update the time step, which is not required for matrix-free operations. 

I have ran tests to measure the performance difference caused by this change on one residual evaluation. The test is ran on Stampede2, one skylake node, compiled with AVX2 instruction set, 48 processors, blockette size 8, test block sizes of 32^3 and 48^3 per processor. I provide a few results:

Base speed: Millions of cells processed per one processor in one second with default residual routines that operate on main arrays.
Blockette speed: Same speed metric, but with cache-blocked residual routines.
Speedup: Blockette speed / base speed

The timing results: 

##### 32^3: 
Base speed: 0.676747811395 
Blockette speed: 1.24815149274 
Speedup: 1.844337686393324 

##### 48^3: 
Base speed: 0.629510355349 
Blockette speed: 1.29542792502 
Speedup: 2.057834178600312

##### 32^3: 
Base speed: 0.676336855482
Blockette speed: 1.21488916591
Speedup: 1.7962782244717301

##### 48^3: 
Base speed: 0.627697421135
Blockette speed: 1.25598801018
Speedup: 2.00094499019755

The addition of this extra memory access causes the speedup to decrease from 1.84 to 1.80 and from 2.06 to 2.00 for test blocks of size 32^3 and 48^3, respectively. This is a very minor decrease in performance. Furthermore, this "slowdown" will only be happening with calls that update the time step, and the main computationally intensive calls (mat-free operations and preconditioner computations) do not update this time step. Therefore, the overall effect of this change on performance will be negligible.

## Type of change

## Testing

Regressions pass. Furthermore, the test in #32 also passes.

## Checklist


joanibal:
I'm going to remove myself as a reviewer because we have already discussed this issue in depth. 

I approve of the changes, but don't want my review to count towards the two required

anilyil:
I was wrong; the bug would not cause adjoint sensitivities to be inaccurate. The intermediate arrays are updated in the call to master in: https://github.com/mdolab/adflow/blob/master/src/adjoint/adjointUtils.F90#L592 The bug would still effect issue #32.

joanibal:
Oops, I thought it wouldn't count it since I removed myself as a reviewer. 
(I approved so pull panda would stop reminding me)

PR #38, opened 2020-04-25
Copy intermediate variables from blockette memory

anilyil:
## Purpose

All of the issues explained above is with the option useblockettes:True, which is the default. Regular, non-blockette residual routines do not have these issues.

As mentioned in issue #36, some intermediate variables that are computed during the residual computation are not copied back to main memory when blockettes are used. This may cause issues when reverse AD routines are called with outdated intermediate variables. This PR modifies the blockette code so that when `updateIntermed` flag is `True`, the blockette residual routine updates all the intermediate variables that are needed by the reverse AD routines. The PR also modifies how `blocketteRes` is called from the NKSolver module so that after an ANK or NK step, the intermediate variables are always up to date.

This does not cause issues with the adjoint solver, because the preconditioner matrix is created before any reverse AD call. In the last step inside the preconditioner computations, master routine is called to update every intermediate variable (see: https://github.com/mdolab/adflow/blob/master/src/adjoint/adjointUtils.F90#L592). As a result, the reverse AD calls made for the adjoint solver and RHS computations are done with updated intermediate variables. However, without the master call in the preconditioner routine, they would be out of date, and this can cause issues down the line, for example, if reverse AD routines are called before the preconditioner computation, the results can be inaccurate before this fix.

Previous issue #36 only mentions spectral radii; however, this is not the full set. Copying spectral radii for Euler simulations is good enough, as it was done in PR #37, but this is not enough for viscous simulations. For these cases, we also need to copy out "speed of sound squared" and "nodal gradient" arrays.

This PR also changes the name of the updateDt flag to updateIntermed because this flag is now used to update all intermediate variables in main memory.
Modified matrix-free residual calls in NK solver so that these intermediate variables are not updated there for performance, but they are updated for NK line search calls.

I have ran the same timing tests I did in PR #37. Refer to PR #37 for the explanation of the data provided:

#### 32^3:

Base speed: 0.680003727937
Blockette speed: 1.24622523872
Speedup: 1.8326741273916345

#### 48^3:

Base speed: 0.627221772801
Blockette speed: 1.29751948271
Speedup: 2.068677362578845

#### 32^3:

Base speed: 0.67284039047
Blockette speed: 1.09666906157
Speedup:  1.6299096741263446

#### 48^3:

Base speed: 0.624760548343
Blockette speed: 1.13382787755
Speedup: 1.81481990269257

Because now we are copying more data from blockette memory, the performance effect of this modification becomes more pronounced. However, the matrix-free residual calls and the residual calls for the FD based approximate preconditioner do not copy this data out, and therefore they have the same performance as previously. 

Matrix-free operations for the NK solver used to copy out the time step dtl previously, with this PR, I modified them so that the matrix-free operations do not copy these intermediate variables, and therefore, they do not see this performance hit. The residual calls for the line search routines all copy back the intermediate variables now. Furthermore, the getResidual call also updates the intermediate variables.

As a result of these changes, after every ANK or NK step, or after a call to getResidual from python level, all intermediate variables required for the reverse AD routines are up to date, and the reverse AD routines return the exact same result as if we used the non-blockette residual routines.

On top, the matrix-free operations for ANK and NK, along with the FD preconditioner operations do not copy these intermediate variables, and they do not experience this slowdown. These are the part of the main computational cost of ADflow. The overall solution time will not be affected by this change as drastically as the results I posted above.

## Type of change

## Testing

The issue mentioned in #32 and #36 was incompletely addressed with #37. This is because we only checked Euler cases previously. @sseraj found that if we do a similar test with a RANS case, the AD output is still wrong. This PR addresses the issue found by him. 

To recreate the problem, the similar approach can be followed that was explained in issue #32 but with a laminar NS or a RANS case: 

```
CFDSolver.getResidual(ap)
dwBar = CFDSolver.getStatePerturbation(314)
wbar = CFDSolver.computeJacobianVectorProductBwd(resBar=dwBar, wDeriv=True)
```

The reverse AD routines that are affected are with the seeds for the residuals and AD outputs for the states. Calling the reverse AD routine just with the functionals alone will not recreate this issue. 

I have tested this fix with the test case from @sseraj with Euler, Laminar NS, RANS, and with scalar, matrix, and upwind discretization. 

The regression tests also pass. 

## Checklist

### DIFF
--- a/src/NKSolver/NKSolvers.F90
+++ b/src/NKSolver/NKSolvers.F90
@@ -312,7 +312,7 @@ contains
     end do spectralLoop4b
 
     ! Evaluate the residual now
-    call computeResidualNK()
+    call computeResidualNK(useUpdateIntermed = .True.)
     call getCurrentResidual(rhoRes, totalRRes)
 
     ! Put everything back
@@ -444,7 +444,7 @@ contains
     ! computeResidualNK subroutine
 
     call setW(wVec)
-    call computeResidualNK()
+    call computeResidualNK(useUpdateIntermed = .False.)
     call setRVec(rVec)
     ! We don't check an error here, so just pass back zero
     ierr = 0
@@ -532,7 +532,7 @@ contains
 
        ! Evaluate the residual before we start and put the residual in
        ! 'g', which is what would be the case after a linesearch.
-       call computeResidualNK()
+       call computeResidualNK(useUpdateIntermed = .False.)
        call setRVec(rVec)
        iter_k = 1
        iter_m = 0
@@ -718,7 +718,7 @@ contains
 
     ! Compute Function:
     call setW(w)
-    call computeResidualNK()
+    call computeResidualNK(useUpdateIntermed = .True.)
     call setRVec(g, flowRes2, turbRes2, gnorm)
 
     nfevals = nfevals + 1
@@ -761,7 +761,7 @@ contains
 
           ! Compute Function
           call setW(w)
-          call computeResidualNK()
+          call computeResidualNK(useUpdateIntermed = .True.)
           call setRVec(g, flowRes2, turbRes2, gnorm)
 
           nfevals = nfevals + 1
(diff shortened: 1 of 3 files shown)


joanibal:
^ that just adds Neils latest changes to the README

PR #41, opened 2020-05-14
updated options

joanibal:
## Purpose
Update the default options of ADflow to make it easier for new users. 

```
'ncycles': 2000
'smoother': 'dadi'
'l2convergence': 1e-8
'nkswitchtol': 1e-5
'anklinresmax': 0.1
'equationtype': 'rans'
'infchangecorrection': True
'useanksolver': True
'nsubiterturb`:3
```
***THIS CHANGE MIGHT BREAK OLD SCRIPTS THAT RELY ON DEFAULT OPTIONS***

## Type of change

- Breaking change (non-backwards-compatible fix or feature)

## Checklist

- [X] I have run unit and regression tests which pass locally with my changes


nwu63:
Just a reminder that I think we should release the current version of the code as v2.1.1, then releasing v2.2.0 once this PR is merged, since this introduces some backwards incompatibility.

nwu63:
Do we want to also change `mgcycle` to `sg` by default? Or do we still want to run multigrid for that? What about `smoother`? Keep it as RK or change to something more commonly used like `dadi`?

anilyil:
I agree with the current set of changes, except blockettes. With everything else being constant, I agree that we should move away from blockettes enabled by default. However, I think the better solution is to: Enable it by default, and have a check on the python layer to ensure that blockettes would be disabled if user wanted anything that would break the blockette formulation. For example, if user wanted to run in unsteady mode, or with a rotating reference frame, we can automatically disable blockettes. These are the main problematic cases that we ran into with blockettes. 

The case for blockettes: As we move onto new CPU architectures, the advantage we get with blockettes will increase. Even currently, we get up to 3x speedup depending on the case. Pre-blockette cost consists of around 50% preconditioner and 50% residual costs. With blockette residual costs being 2-3 times lower, with blockettes, any improvement with the preconditioner would yield huge speedup values. Without blockettes, these would be marginal. See issue #49 for a promising update in preconditioner performance.

I was originally for moving away from blockettes by default, but I think we should keep them enabled by default. My main point is: yeah it is more difficult to maintain and was not tested well, we can keep debugging and add tests that cover this stuff to keep it as the default. I believe the future of computing will involve any serious code having cache-blocking functionality, and therefore, I think every effort going towards debugging blockettes will be useful in the long run.

Besides these, additional changes I would recommend:
```
'ncycles': 2000, # the default 500 is not enough for any practical RANS case
'smoother': 'dadi', # with the dadi smoother, the partitioning code can produce a slightly different domain partitioning, where longer lines are preserved in the off-wall direction. I think this has been beneficial for all implicit solvers in ADflow (ANK, NK), so I think we should switch the default to DADI. We can take a look at the partitioning in a separate issue.
'l2convergence': 1e-8, # 1e-6 is a bit high for any transonic rans case. I think everyone already sets this in their runscripts, but 1e-8 would give a better baseline to work with.
'nkswitchtol': 1e-5, # the default 2.5e-4 is too high for any practical transonic case, or any case with separation
'anklinresmax': 0.1, # we added this after a suggestion by @camader, the default is 0.9. this option proved to be *extremely* useful for every case I worked on. I always set it to 0.1, this is basically over-writing the CFL upper limit. I am for this change because of the additional robustness we would get with it. I will document this option separately in the adflow docs, and possibly add it to the solver docs.
```

I would say the list above is an "inclusive" one rather than a minimal one. That is; I tried to comment on all of the default options I could think of. It doesn't mean I think we should change all of them at once, but I think it is good to have a discussion on them.

camader:
I am fine with the options changes listed in this PR. With respect to blockettes, I am fine either way, but we leave them on by default, I agree that adding in the checks is a good idea.

nwu63:
Do we think the tests have stabilized enough to get this one going? Obviously we need to make a new release before merging this.

joanibal:
OK lets leave the blockettes on by default.
The final set of options are 

```
'ncycles': 2000
'smoother': 'dadi'
'l2convergence': 1e-8
'nkswitchtol': 1e-5
'anklinresmax': 0.1
'equationtype': 'rans'
'infchangecorrection': True
'useanksolver': True
```

joanibal:
When #43 is merged, then I'll update this PR

sseraj:
Maybe we should add `'usenksolver': True`?

joanibal:
The NK solver is not very robust and so I think it should stay off by default

joanibal:
@nwu63  this is ready for review again

sseraj:
Nice! Could you also increment the version number?

anilyil:
@joanibal can you also increase `nsubiterturb` if we did not already? I thought we were going to do that but I am not 100% sure.

marcomangano:
> OK lets leave the blockettes on by default.
> The final set of options are
> 
> ```
> 'ncycles': 2000
> 'smoother': 'dadi'
> 'l2convergence': 1e-8
> 'nkswitchtol': 1e-5
> 'anklinresmax': 0.1
> 'equationtype': 'rans'
> 'infchangecorrection': True
> 'useanksolver': True
> ```


@joanibal once you are done can you also copy this on the PR description, just for quicker reference?

sseraj:
The docs for `nSubiterTurb` should be updated as well now that the default is 3.

anilyil (dismissed):

sseraj (approved):

nwu63:
We've apparently hit our rate limit again, so if @anilyil approves it I will merge.

anilyil (approved):

Issue #44, opened 2020-05-26
getFreeStreamResidual method changes functionals for viscous equation types


joanibal:
## Type of issue
What types of issue is it?
This is a bug  

## Description
The pyADFlow function `getFreestreamRes` is causing a state change for viscous equation types ('laminar ns', 'rans'). 

This is a problem since many of our old regression test use the helper function standard test, which calls `getFreestreamRes` and then evaluates the functionals. 
This means that functional values is some reference files are wrong.

### Steps to reproduce issue

call evalFuntions
call getFreestreamRes
call evalFunctions # this will give you a different answer 

see [runscript](https://gist.github.com/joanibal/1daf08fb4738f52fee122afb0feb6e5d)
…

### Expected behavior
…

call evalFuntions
call getFreestreamRes
call evalFunctions # this will give you the same answer 

### Code version (if relevant)
**Python version:** … 2.7

**External dependencies:** …

**Internal packages:** …
ADFlow 2.1.0

shamsheersc19:
@joanibal Any updates on this?

PR #46, opened 2020-05-29
bug fix for use of  getFreeStreamRes with tests

joanibal:
## Purpose
Fix for issue #44 

## Type of change
What types of change is it?

## Testing
**TESTING** not retrained to work with fix. 
I made this a draft to get feedback before I do the retaining


joanibal:
The test are going to fail, because the viscous tests had a bug before.

anilyil:
Your diagnosis is correct, and this does fix this issue. However, I did a bit of digging, and the issue can also be fixed in a few other ways. The root cause of this issue is due to the viscous stresses at the walls being over-written during the free-stream residual evaluation. If we can avoid over-writing these values, then we would not need the additional residual evaluation.

With the blockette residual routines, we already have an option to avoid copying these values during the viscous flux calculations. For this, one bug needs to be fixed first:
https://github.com/mdolab/adflow/blob/master/src/NKSolver/blockette.F90#L5523-L5526 This if check used for the optional variable `storeWallTensor` is wrong, since we would need to grab the actual value of this if it is present, rather than setting the flag to true. The fixed version would look like this:
```
    storeWall = .False.
    if (present(storeWallTensor)) then
       storeWall = storeWallTensor
    end if
```
Furthermore, this check only copies out the viscous stresses, and not the heat fluxes. @joanibal if you have additional cost functions that use the heat flux, we would need to add these here as well.

Once these are fixed, the blockette version can properly avoid copying out these values.

We also need to add the capability to skip copying the viscous stresses in the "regular" viscous flux calculations. These are done in the lines:
https://github.com/mdolab/adflow/blob/master/src/solver/fluxes.F90#L2824-L2854 , 
https://github.com/mdolab/adflow/blob/master/src/solver/fluxes.F90#L3117-L3148 , and
https://github.com/mdolab/adflow/blob/master/src/solver/fluxes.F90#L3413-L3447 .
We can add an optional variable similar to the blockette version so that the copy is skipped. 

There are two options to add this:
1. We can simply add the if check that uses the optional input variable, but this would also require re-running tapenade since the code would be modified. The actual computation in the AD output will not change, but we will get this logical added to the AD output.
2. We can add the if check and use a precompiler directive that disables the if check when we are running tapenade. This way, we won't need to re-run tapenade and generate the updated AD code.

So bottom line, I think we have 3 options:

1. Keep the PR as is. This fixes the issue at the cost of an extra residual evaluation. 
2. Add the functionality to skip copying the viscous stresses in the `viscousFlux` subroutine in https://github.com/mdolab/adflow/blob/master/src/solver/fluxes.F90 . Re-run tapenade and update the AD code. Remove the extra residual evaluation introduced in this PR. Change the NKresidual call to a `call blocketteRes(useUpdateIntermed = .false., useStoreWall=.false.)` or add the optional variable `useStoreWall=.False.` to the NK residual. Re-run tapenade and update the AD routines.
3. Do exactly as it is done in Step 2, except add the if checks to the `viscousFlux` routine using precompiler directives and have these checks disabled when we are running tapenade, so we don't need to update the AD code.

I think we should either pick option 1 or 3. Option 1 is good because it is simple, yet, it hides the underlying issue. Option 3 is good because everything is done explicitly, but its more work and more complicated code (relatively).

Regardless of what option we pick, we should fix:
1. The if check I mentioned related to the `storeWallTensor` optional variable check in blockette.
2. Also copy out the heat fluxes when we run with blockette code. Currently, I dont think we have a cost function that uses this, but it seems like the heat fluxes can be an output in the `heatFluxes` method in https://github.com/mdolab/adflow/blob/master/src/warping/getForces.F90 Not copying out these heat fluxes from blockette residuals will likely cause issues for your aerothermal stuff @joanibal .

joanibal:
I'd say I'm a strong advocate for option 1. 
Correct me if I'm wrong but the cost a single residual evaluation is negligible compared to the cost of a solve.
I don't follow what you are saying about " copying out these heat fluxes from blockette residuals" so we should have a chat sometime.

anilyil:
I agree, cost of a single residual is very small. However, I don't want to add a fwd residual routine every time we run into a similar issue to avoid bloating the code.

I think option 3 would explicitly point developers to the correct place in the code if something needs debugging down the line, vs. just adding a residual evaluation is more like treating the symptom. Then again, option 1 makes sense because the code is so simple... I am not sure the best way here.

Regarding the heat fluxes; my bad, they are being copied out. I missed the lines where q is copied out along with tau.

joanibal:
I think simpler is better. I'm not interested in investing the time to solve this the more complicated way. We don't even really have the test coverage to say that the modifications to the core residual routines to fix this problem won't create other ones.

If you can live with the simpler fix, I'll submit this as an actual pull request.

anilyil:
Okay that makes sense. Simple version it is. Can you please:
1. Add comments on why we need the additional residual call. Like directly say what variables are out of date, so it will be easier in the future if we get back to this. Maybe mention this PR with the number?
2. Can you fix the issue I mentioned above with the storeWall check in blockettes. The line https://github.com/mdolab/adflow/blob/master/src/NKSolver/blockette.F90#L5525 should be replaced with `storeWall = storeWallTensor`.

nwu63:
What's the status here? Is this PR ready?

anilyil:
Looks good to me. Seems like @joanibal addressed my comments. I think he needed this stuff for some of the testflo migration, so I think he would know better if we want to merge this now or not.

anilyil (dismissed):

joanibal:
Yeah it is all good to go. 
just have to merge the master.

joanibal:
@anilyil 
@Xiaosong2105 

It is ready for review

anilyil (approved):
thanks for the fix @joanibal

nwu63 (approved):

PR #56, opened 2020-06-12
setup.py install

joanibal:
## Purpose
Change the structure of the repo to use the `setup.py` method of installation. 
This will make it easier to install environment specific packages and specify dependencies. 

One can install the repo now by doing
`make` (if applicable) and `pip install .`

## Type of change
- Code style update (formatting, renaming)
- Refactoring (no functional changes, no API changes)

## Checklist


nwu63:
The complex tests need to have the imports changed, mirroring what you did for the real tests.

Actually, both real and complex imports should be changed:
- no sys.path modification needed
- I think you can directly do `from adflow import ADFLOW` and `from adflow import ADFLOW_C`, since they should be provided by the `__init__.py` file.
- the warning about import strategy can be finally removed, since we will now be importing things "correctly"

nwu63:
Correct, I think we should do optional dependencies for idwarp eventually, but it's okay for now. Once we move over to testflo we can also skip tests based on what's installed.

joanibal:
@nwu63 Are you happy with the import strategy now?

nwu63:
Yes I am.

joanibal:
could you approve the review then?

PR #57, opened 2020-06-21
Updated docs

nwu63:
## Purpose
Updated path for autodoc so the API can be rendered properly

## Type of change
What types of change is it?

- Documentation update

### DIFF
--- a/doc/pyADflow.rst
+++ b/doc/pyADflow.rst
@@ -4,5 +4,5 @@ pyADflow
 ---------
 .. currentmodule:: python.pyADflow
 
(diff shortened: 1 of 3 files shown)


doc/pyADflow.rst
```diff
@@ -4,5 +4,5 @@ pyADflow
 ---------
 .. currentmodule:: python.pyADflow
```
sseraj, line 5: Should this also be changed?

nwu63:
@sseraj should be good to go now.

Issue #50, opened 2020-06-04
Migrate the MPhys wrapper and move the code from the MPhys repo to here


anilyil:
## Type of issue

- New feature (non-breaking change which adds functionality)

## Description
The current om_adflow wrapper for OpenMDAO is slowly becoming depreciated as we are figuring out more details about the MPhys wrapper, which achieves a similar functionality in a more general way. We will remove the current om_adflow wrapper for good and move the MPhys wrapper from the MPhys repository to here once the development is mature enough.

PR #60, opened 2020-06-23
Update test_adjoint_compare.py

DavidAnderegg:
## Purpose
Since the package name was updated, this should be updated accordingly.

## Type of change
What types of change is it?

## Testing
Run the file "run_om_reg_tests.py".

## Checklist

### DIFF
--- a/reg_tests/om_tests/test_adjoint_compare.py
+++ b/reg_tests/om_tests/test_adjoint_compare.py
@@ -24,7 +24,7 @@ from pyspline import Curve
 sys.path.append(os.path.abspath('../../'))
 
 sys.path.append(os.path.abspath('../'))
-from python.om_utils import get_dvs_and_cons
+from adflow.om_utils import get_dvs_and_cons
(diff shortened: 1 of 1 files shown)


anilyil:
@DavidAnderegg, we do not actively maintain the OpenMDAO wrapper that is in this repo. We are working on another version with a few improvements. Feel free to keep using this, but we suggest you perform your own testing to make sure this wrapper works. We will release the new wrapper once it is more or less finalized and ready for general use.

DavidAnderegg:
@anilyil, Thank you for your reply.

Do you have any idea when that release will happen? I am eager to test it out.

anilyil:
Thanks for the interest @DavidAnderegg. We will release it possibly some time over the summer. The updated ADflow wrapper inherited a lot of the code from the current wrapper. Hopefully the transition to the new version will be smooth. Keep an eye out for issue #50. We will update that with the progress related to this.

PR #61, opened 2020-06-23
Updated gitignore, and added references

anilyil:
## Purpose

The gitignore file was updated according to the recent changes in folder structure. I also added the ADflow, ANK, and adjoint papers to the readme.


nwu63:
IMO there should be a dedicated citation page in the docs, rather than having all the information in the README.

anilyil:
I think having the citations in readme is not just for people who will use the code and cite the papers, it is also for publicizing the papers. And not everyone will read the docs

nwu63:
Not sure how effective it's going to be at publicizing papers, that probably should be done elsewhere like on research gate or something. I don't think people browse GitHub repos to read papers -- people are here to look for code. And the argument that "not everyone will read the docs" isn't a very good one either, I don't think we want to cater to that.

In any case, I still think we should have a citation page, where we supply the full BibTeX entry to people.

FFR if you are editing these text files you can suppress Travis by adding `[skip travis]` at the beginning of your commit message.

anilyil:
I will create the preprint for the ADflow paper and link to this PR. We can merge after I add the url for the preprint.

anilyil:
You can check out how the readme looks like here: https://github.com/anilyil/adflow/blob/patch/README.md In the last commit, I added the correct URL for the adflow paper preprint, tweaked the readme text a bit, and added the figure.

PR #62, opened 2020-06-23
Changed travis to run on 2 procs

nwu63:
## Purpose
I changed Travis to run tests using 2 processors, since that's what's available. Oversubscribing doesn't make sense and I don't think this reduces test coverage in any way.


nwu63:
The reason py2 failed is because
- default options have key in all lower case, but for tests 16-18 the updated options had key in mixed case, so the final options dictionary had both `gridfile` and `gridFile` entries
- in py2 the dictionary order is different than py3, so by chance py3 took the `gridFile` option when py2 took the `gridfile` option

I will push a fix where all updated options also have their names in lower case. In the future, maybe we should implement a `CaseInsenstiveDict` or something for the options?

PR #64, opened 2020-07-01
Add Doxygen

nwu63:
## Purpose
I added doxygen build to our RTD builds.
- updated `conf.py` to build Doxygen on RTD. The approach is taken from [here](https://breathe.readthedocs.io/en/latest/readthedocs.html)
- minor formatting adjustment to the CGNS grid module so that Doxygen doesn't segfault
- updated doxyfile with `doxygen -u`

## Type of change
What types of change is it?

- Documentation update


nwu63:
I see that the build time on RTD is significantly longer now, but the Doxygen stuff is generated correctly and can be accessed from the index.

eirikurj:
Merging since we have two approvals.

PR #79, opened 2020-09-19
added note about HPC build options

marcomangano:
## Purpose
I added a note to the installation documentation after a feedback from collaborators from BYU. For some HPC configurations the specific CPU architecture of the compute nodes needs to be specified for `FF90_FLAGS`. 
The general recommendation remain to contact the local HPC team for hardware related issues.

## Type of change
- Documentation update

## Checklist

- [x] I have built the documentation locally on my machine


sseraj (dismissed):
The build should actually work because this is a public repo. I'm approving now since you only changed the docs.

nwu63 (dismissed):

doc/install.rst
```diff
@@ -47,6 +47,14 @@ To configure manually, open ``config/config.mk`` and modify options as necessary
 It is most likely that you need to modify the ``CGNS_INCLUDE_FLAGS`` and the ``CGNS_LINKER_FLAGS`` variables.
 After changes to the configuration file, run ``make clean`` before attempting a new build.
 
+.. NOTE::
+
+    Some HPC architectures might require modifications on additional flags in the ``config.mk`` file. Users have documented cases of abrupt ADflow crashes, despite a successful build, right after the job submission, together with other issues with interactive jobs. For example, problems were reported for a cluster where login nodes use newer Cascade Lake CPUs while compute nodes are based on older Sandy Bridge CPUs.
+
+    We recommend to contact your local HPC team to address these hardware specific issues. To provide an example fix, the issue on the above mentioned HPC was fixed using the following ``FF90_FLAGS`` option instead of the default one::
+
+        FF90_FLAGS = -DHAS_ISNAN  -fPIC -fdefault-real-8 -fdefault-double-8 -g  -O3  -march=sandybridge -mtune=broadwell -ffast-math
```
eirikurj, line 56: Did this suggestion come from their support team? To me it seems strange to set `-march=sandybridge` but then optimize for an older broadwell. This may depend on their compiler version but specifying `-march=sandybridge` should set the tuning to sandybridge, which is what you want generally, so usually there is no need to specify `-mtune`.
Furthermore, if they have Cascade Lake, you should probably use `-march=cascadelake`, if the compiler supports it, but this may make the binary non-compatible with the login node (should be fine since they should not run it there anyways).

To future proof this, I suggest adjusting the above text and remove all "standard" flags that are there from the `config.mk` (in case it change) and only include the relevant extra `-march` and `-mtune` flags e.g.,
```
FF90_FLAGS = <normal-flags> -march=HPC-ARCH -mtune=HPC-ARCH
```
or similar.

eirikurj (changes requested):

doc/install.rst
marcomangano: Sorry for the late reply! Yes this came from their HPC team, but I understand your point. What is the recommended action item then @eirikurj ? Shall we modify and keep this note and/or edit the `config.mk` file?

eirikurj: I think adding a note is fine, but I am not sure we need to change the config files.
As I suggested I would keep this more general than what is stated there since people have many compiler versions that do support different levels of architectures. I would add something more general like what is shown below. 

```
NOTE:
Care must be taken when compiling on HPC systems as some systems do not have a homogeneous CPU architecture across all nodes.
For example, the login node architecture may differ from the compute nodes architecture. Compiling the code for a specific login node architecture may result in unexpected crashes on the compute nodes if incompatable (newer) architecture is used.
To specify the architecture of the compute node you can append the `-march=HPC-ARCH` flag to your `config.mk`. Furthermore, if architecture optimization is desired for a specific architecture, one can add the `-mtune=HPC-ARCH` flag also. This however is rarely needed. An example of an updated line in the config file is, 

`FF90_FLAGS = <normal-flags> -march=HPC-ARCH -mtune=HPC-ARCH`

Note that the `HPC-ARCH` should be replaced with the correct compute node architecture. This name is compiler specific and depends on the compiler version also.
For example, if the login nodes use newer Cascade Lake CPUs while the compute nodes are based on older Sandy Bridge CPUs it may be necessary to set `-march=sandybridge` in your `config.mk` file or

`FF90_FLAGS = <normal-flags> -march=sandybridge`
```
or something similar. Feel free to adjust this.

marcomangano: @eirikurj thanks a lot, that makes much more sense. I edited the note according to your comments, please see the latest commit

nwu63: Now that the note is much more general, does it make sense to have this in the general installation page rather that for ADflow? Do you see this being useful for other packages like IDwarp for example?

eirikurj (approved):

nwu63 (approved):

PR #84, opened 2020-10-14
Updated nprocs for tests

nwu63:
## Purpose
After some confusion with testflo, it turns out `-n` specifies the number of tests to run concurrently. Therefore, I am setting it to `1` for Travis. I am also changing the `N_PROCS` parameter to 2 since it's not really necessary to run MPI tests with 4 procs, and this would ensure some sort of consistency between Travis and the build bot.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

### DIFF
--- a/tests/reg_tests/test_adjoint.py
+++ b/tests/reg_tests/test_adjoint.py
@@ -87,7 +87,6 @@ test_params = [
         "ref_file": "adjoint_euler_scalar_jst_tut_wing.json",
         "aero_prob": ap_tutorial_wing,
         "evalFuncs": ["cl", "cd"],
-        "N_PROCS": 1,
     },
     # Tutorial scalar JST
     {
@@ -176,7 +175,7 @@ class TestAdjoint(test_objects.RegTest):
     based on old regression tests 12, and 14
     """
 
-    N_PROCS = 4
+    N_PROCS = 2
 
     options = None
     ap = None
@@ -233,7 +232,7 @@ class TestCmplxStep(unittest.TestCase):
     based on old regression tests_cs 12, and 14
     """
 
-    N_PROCS = 4
+    N_PROCS = 2
 
     h = 1e-40
 
--- a/tests/reg_tests/test_adjoint.py
+++ b/tests/reg_tests/test_adjoint.py
@@ -2,7 +2,6 @@
 import unittest
 import numpy
 import os
-import sys
 import copy
 from collections import defaultdict
 from parameterized import parameterized_class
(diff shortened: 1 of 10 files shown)


tests/reg_tests/test_adjoint.py
```diff
@@ -87,7 +87,6 @@ def span(val, geo):
         "ref_file": "adjoint_euler_scalar_jst_tut_wing.json",
         "aero_prob": ap_tutorial_wing,
         "evalFuncs": ["cl", "cd"],
-        "N_PROCS": 1,
```
nwu63, line 90: @joanibal can you explain what this does? I removed it here because I didn't see it getting used anywhere, but can add it back if it does something.

joanibal: That's for the single core test. It overrides the attributes which specifies the number of cores from the base class used by parametrized. 

If we don't want a single core test than we can get rid of it.

joanibal (changes requested):
See my comment. I think we should probably keep the single core test. It could be useful for debugging multi core problems

tests/reg_tests/test_adjoint.py
nwu63: Oh I see, those become class attributes when the parameterized class is instantiated? Did you check that this actually works, i.e. testflo calls it with 1 proc? That's pretty neat

joanibal: Exactly, the parameterized_class decorator is actually really simple. 
```
    def decorator(base_class):
        test_class_module = sys.modules[base_class.__module__].__dict__
        for idx, input_dict in enumerate(input_dicts):
            test_class_dict = dict(base_class.__dict__)
            test_class_dict.update(input_dict)

            name = class_name_func(base_class, idx, input_dict)

            test_class_module[name] = type(name, (base_class, ), test_class_dict)
```


It works, I even rechecked before commenting haha

joanibal (dismissed):

tests/reg_tests/test_adjoint.py
nwu63: Very cool! I recall `paraeterized` saying unittest only supports `@parameterized.expand`, but I guess it works here because we're not actually using unittest, i.e. calling `python test_adjoint.py`, and testflo supports these decorators? Pretty neat stuff, I'll keep this in mind when I parameterize pyoptsparse tests.

eirikurj (approved):

sseraj (commented):
The complex tests are failing marginally again, but this time on Ubuntu 18.

nwu63:
> The complex tests are failing marginally again, but this time on Ubuntu 18.

Yeah I just pushed up a fix. I am baffled why the build bot is passing... the inconsistency is concerning, esp. since both are using Docker and that should resolve these issues.

tests/reg_tests/test_adjoint.py
joanibal: The tests will run just fine with `python test_adjoint.py	 (I just ran it to check). I'm not entirely sure why though.

The docs for parametrized are kinda confusing about that. They say only @parameterized.exapand will work with unittest but then they show the @paramertized_class working...

sseraj (approved):

joanibal (approved):

PR #92, opened 2020-11-15
Updated displacement file handling

nwu63:
## Purpose
This PR fixes a minor logic in applying displacement files. Now, even if the point sets are up to date, if the displacements are nonzero they are still applied. I also made minor adjustments to the docstrings.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -413,20 +413,27 @@ class ADFLOW(AeroSolver):
 
     def setDisplacements(self, aeroProblem, dispFile):
         """
-        This function allows the user to perform aerodyanmic
+        This function allows the user to perform aerodynamic
         analysis/optimization while using a fixed set of displacements
         computed from a previous structural analysis. Essentially this
-        allows the jig shape to designed, but performing anlysis on
-        the flying shape. Note that the fixed set of displacements do
-        not affect the sensitivities.
+        allows the jig shape to designed, but performing analysis on
+        the flying shape.
 
         Parameters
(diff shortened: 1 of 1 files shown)


Xiaosong2105 (approved):
Good catch. I have approved this PR.
I am thinking if every time we finish updating the point sets with displacement, we can reset the displacement to zeros.
Thus, the logic in the modified file can be simplified by only considering the current displacement is zero or not.

nwu63:
That is a good point, I'm not sure how this will work with optimization. No idea if the displacements should be zeroed out, or applied every time.

marcomangano:
That's a nice catch indeed! I am wondering how this affects the MDA/O results in terms of speed and accuracy

eirikurj (approved):
Looks good.
A note on usage, the `setDisplacements` function is rarely used and is not used in an aerostructural MDA e.g., `pyaerostructure`. This function is useful for example in an ASO when you want to deflect the aerodynamic shape depending on some structural loading obtained from a separate structural analysis. This would remain fixed throughout the optimization.
As for resetting it to `None` or zero, I dont see that being useful here because these displacements are intended to stay fixed throughout.
A minor comment, while its very likely that you will have a DVGeo object, it should probably be possible to set the displacements without requiring DVGeo object present. This could be done later though as I think 99% of the time you have a DVGeo object present.

PR #102, opened 2021-01-08
Options documentation and refactoring

### LINKED ISSUE #43: Document and update adflow options

sseraj:
## Purpose
This PR adds documentation for all ADflow options. (Closes #43)

The options table is generated automatically from the default options dictionary in pyADflow and the descriptions in options.yaml. Doing this required modifying pyADflow to use ``getOption`` and ``setOption`` from BaseSolver in baseclasses.

There are two changes in this PR that may break backward compatibility with ADflow run scripts:
1) String-type option values are now case sensitive and some of the acceptable string values have changed. This will most commonly affect `equationType`. For example, you must use `"RANS"` instead of `"rans"`.
2) ``autosolveretry``, ``autoadjointretry``, and ``nkcfl0`` are now deprecated. These options have had no function for years but may still be in old scripts.

The full list of affected string-type options is:
```
solutionPrecision
gridPrecision
solutionPrecisionSurface
gridPrecisionSurface
discretization
coarseDiscretization
limiter
smoother
equationType
equationMode
flowType
turbulenceModel
turbulenceOrder
turbulenceProduction
eulerWallTreatment
viscWallTreatment
resAveraging
oversetUpdateMode
timeIntegrationScheme
NKLS
adjointSolver
preconditionerSide
matrixOrdering
globalPreconditioner
localPreconditioner
```

These changes will break the pyAeroStructure tests. @eirikurj will fix the options either for the current tests or with the [testflo update](https://github.com/mdolab/pyaerostructure/pull/32). I will submit a PR to fix the tutorial run scripts after this is merged.

Once this and #41 are merged, we will make a new release and make the backward incompatible changes clear in the release notes.

The formatting of options table itself is not the best, but we can address this in the future. See [here](https://github.com/mdolab/sphinx_mdolab_theme/issues/3) for a possible solution.

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
See the [BaseSolver tests](https://github.com/mdolab/baseclasses/blob/master/tests/test_BaseSolver.py) for how ``getOption`` and ``setOption`` are tested.

## Checklist


nwu63:
Do we want to address #43 before incrementing the minor version/making a release?

sseraj:
I think #43 has been addressed? Is there something missing from this PR?

nwu63:
I guess I'm referring more to #41, whether we want to wait until that's resolved before releasing a new minor version.

sseraj:
I'm leaning towards separate releases, but it really depends on quickly we can merge #41 (or a new PR) after this one. What do you think @joanibal?

joanibal:
I suggest moving forward with this PR. There hasn't been much activity on #41 for a while. 

I'm not sure what the next steps are for #41, maybe we could talk about it at a maintenance meeting?

adflow/pyADflow.py
```diff
@@ -5102,17 +5051,14 @@ def _getSpecialOptionLists(self):
         """
         Lists of special options
         """
-        # These "ignore_options" are NOT actually', ignored, rather,
```
joanibal, line 5105: >"ignore_options" are NOT actually', ignored

oof, glad we changed that variable name

joanibal (approved):

adflow/pyADflow.py
```diff
@@ -4599,34 +4552,34 @@ def _getDefOptions(self):
             'cflcoarse':[float, 1.0],
             'mgcycle':[str, '3w'],
             'mgstartlevel':[int, -1],
-            'resaveraging':[str,'alternateresaveraging'],
+            'resaveraging':[str, ['alternateresaveraging', 'noresaveraging', 'alwaysresaveraging']],
```
nwu63, line 4555: Are you planning to update these names?

nwu63:
So I just want to highlight again the fact that this PR is proposing to make all option values that are strings to be lower case for consistency. I'm still not 100% convinced since it will reduce some readability, but I also don't feel strongly either way. But everyone should understand the consequences and voice your opinions accordingly.

Alternatively, with a bit more work we can implement something in baseclasses (say a set of option names) where the values should be case insensitive. Then for get/set options they can be cast to lower and made consistent, but the user would be free to input with whatever case they like. I'm not necessarily advocating for this approach, but if we want to preserve the freedom then we definitely want some sort of systematic approach like this rather than an ad-hoc `str.lower()` in random places.

adflow/pyADflow.py
sseraj: I decided not to for now to keep the backward incompatible changes to a minimum. I could do it though if you think it's worth changing.

nwu63: I personally think if we are to make backwards incompatible changes we might as well do them at once. That's why I also asked about #41, I think it's more work to announce/release changes multiple times rather than all at once. Up to you guys though since I'm not even a maintainer for adflow.

joanibal: Good point about "squashing" the backwards incompatible changes into one commit. 

I was still in the mindset of 1 change per PR. 
Is there a way to still do one PR per feature and only make one update on the master branch?
Could we collect some backwards incompatible changes in a separate branch?

joanibal:
I'm curious about the requirement for lowercase options too. 
I assumed it had something to do with the automatic documentation process.

adflow/pyADflow.py
nwu63: I was not thinking about one commit vs. multiple on the master branch, IMO it's fine to have multiple backwards incompatible changes on the master branch. It's more a matter of when to do the releases since my mindset is that that's when changes are formalized, listed and communicated to end users. It's generally not a good idea to just pull down the latest changes on master, and checking out a tagged release should help remind users to read the release notes.

nwu63:
> I'm curious about the requirement for lowercase options too.
> I assumed it had something to do with the automatic documentation process.

Nope it's not due to that, we just wanted to formalize a list of strings which are accepted as possible choices. I do think it's more explicit and less error-prone to accept only one capitalization rather than all possible capitalizations, but whether that's camelCase or lower case is a matter of readability vs. ambiguity.

So I guess there are three options here
1. force exact string matching, lower case only
2. force exact string matching, "reasonable" capitalization is used
3. arbitrary capitalization is accepted

See with a GUI we wouldn't need to have any of this debate :P anyways my ranked choices are 213 for this.

joanibal:
Oh I see

Hmm, my preference is 132, since other repos have already been updated with option 1. At the end of the day this isn't that big a deal for me.

adflow/pyADflow.py
joanibal: How would you all feel about making these changes, but instead of causing an error, incorrect capitalization leads to a deprecation warning. 

Then we can Merge this PR but avoid incrementing the major version number.

nwu63:
I don't feel super strongly about this either although if we want to settle on a satisfactory choice now is the time to do it. I'd rather change everything now than revisit this in a year's time.

sseraj:
My preference is also 132, but mainly because I think all lowercase is easiest to remember. I have no problem changing all the scripts again if we choose option 2. I will also gladly work on option 3. I agree that the important thing is that we make an intentional choice now rather than flip-flopping in the future. Let's see what @eirikurj and @anilyil think.

eirikurj:
I do realize that from a coding standpoint that its easier to set and expect everything to be lowercase  in order to avoid having to remember how the option is exactly specified.
Regardless of the option we choose, I also think its not a big ask on the user to resort to the options documentation to check how the options is defined and should be used.
Overall, I dont feel that strongly about this, but my preference is to favor readability, so I would prefer an overall camelCase strategy for both the option name and value.
However, using a strict camelCase rule is probably not going to be possible since there will be some special cases such as `Euler`, and acronyms should probably be capitalized, such as `RANS` or `GMRES`.
To avoid these issues and maintain backwards compatibility we should favor 3.
However, if we want to enforce specific case for names and values and are set on breaking backwards compatibility, I think we should go for 2 (with some exceptions) due to readability.
To close this discussion I think we should meet and discuss pros and cons of each strategy.

adflow/pyADflow.py
```diff
@@ -4516,30 +4481,17 @@ def setOption(self, name, value):
...
             'gridfile':[str, 'default.cgns'],
-            'restartfile':[object, None],
+            'restartfile':[(str, list, type(None)), None],
 
-            # Surface definition parameters:
-            'meshsurfacefamily':[object, None],
-            'designsurfacefamily':[object, None],
-            'closedsurfacefamilies':[object, None],
+            # Surface definition parameters
+            'meshsurfacefamily':[(str, type(None)), None],
```
nwu63, line 4492: I thought these options take in either None (default), str, or list of string? Do we need to add `list` to the possible types?

sseraj: Maybe @anilyil can clarify this, but for what I gather:
Options with `family` only accept a string or None.
Options with `families` only accept a list or None.

adflow/pyADflow.py
```diff
@@ -4599,34 +4552,34 @@ def _getDefOptions(self):
...
+            # Overset Parameters
             'nearwalldist':[float, 0.1],
             'backgroundvolscale':[float, 1.0],
             'oversetprojtol':[float, 1e-12],
             'overlapfactor':[float, 0.9],
             'oversetloadbalance':[bool, True],
             'debugzipper':[bool, False],
-            'zippersurfacefamily':[object, None],
+            'zippersurfacefamily':[(str, type(None)), None],
             'cutcallback':[object, None],
```
nwu63, line 4569: this could be replaced by `(types.FunctionType, types.BuiltinFunctionType)` I think. Not super important but we are expecting a callable here right?

adflow/pyADflow.py
nwu63: Yes I think you are right, options that behave like I described (such as `restartFile`) has been updated already.

adflow/pyADflow.py
sseraj: Good call. I will update this. I'm inclined to not include `types.BuiltinFunctionType` though because it really should not be a built-in function.

nwu63: Sure, I wasn't sure what types of functions gets passed in.

adflow/pyADflow.py
```diff
@@ -4514,15 +4514,15 @@ def _getDefaultOptions():
             # Physics Parameters
             'discretization':[str, ['central plus scalar dissipation', 'central plus matrix dissipation', 'upwind']],
             'coarseDiscretization':[str, ['central plus scalar dissipation', 'central plus matrix dissipation', 'upwind']],
-            'limiter':[str, ['vanalbeda', 'minmod', 'nolimiter']],
-            'smoother':[str, ['runge kutta', 'dadi']],
-            'equationType': [str, ['euler', 'laminar ns', 'rans']],
+            'limiter':[str, ['vanAlbada', 'minmod', 'noLimiter']],
```
nwu63, line 4517: Since we are already breaking some backwards compatibility, do we want to "fully" update the options? e.g. `van Albada`, `no limiter` and `k-omega`?

nwu63: I'm asking since I see a lot of options already have spaces, like `Runge Kutta` instead of `rungeKutta`.

sseraj: I don't see why not at this point. Most of the string options do use spaces so I think it would be nice to make it consistent. What about dashes? We don't use those anywhere. Is it better to stick with spaces?

nwu63: I think we should use dashes if that's what done in the literature, e.g. `Wilcox k-omega`, but if it's just two words then we should split with spaces. Overall, I think it's fine to keep option names as single words with camelCase, but the values could be expanded to have spaces (or dashes if appropriate). What do you guys think @eirikurj @anily @joanibal

sseraj: I added them for now, but it's easy enough to change back.

joanibal:
Is this ready to be reviewed again?

sseraj:
Yes, it is!

joanibal (approved):

anilyil:
Thanks for this work @sseraj! Just to be clear; the users will need to update the options that have string values if they were modified in this PR correct? How will we communicate this change with users? In release notes? 

Besides that, the PR looks good to me.

sseraj:
Yes, the users will have to update the string values. To communicate this, I am planning on listing the names of the affected string options in the release notes and linking to the options table for the new values. I can also add the list to the PR description.

adflow/pyADflow.py
```diff
@@ -1897,8 +1888,8 @@ def solveSep(self, aeroProblem, sepStar, nIter=10, alpha0=None,
         # Name of function to use
         funcName = '%s_%s'%(ap.name, sepName)
 
-        if not self.getOption('rkreset') and self.getOption('usenksolve'):
-            ADFLOWWarning("nRKReset option is not set. It is usually necessary "
+        if not self.getOption('rkreset') and self.getOption('usenksolver'):
+            ADFLOWWarning("RKReset option is not set. It is usually necessary "
```
nwu63, line 1892: Is this still true with ANK?

adflow/pyADflow.py
```diff
@@ -4516,259 +4482,244 @@ def setOption(self, name, value):
...
-            'useapproxwalldistance':[bool, True],
-            'eulerwalltreatment':[str, 'linear pressure extrapolation'],
-            'viscwalltreatment':[str, 'constant pressure extrapolation'],
-            'dissipationscalingexponent':[float, 0.67],
+            'discretization':[str, ['central plus scalar dissipation', 'central plus matrix dissipation', 'upwind']],
+            'coarseDiscretization':[str, ['central plus scalar dissipation', 'central plus matrix dissipation', 'upwind']],
+            'limiter':[str, ['van Albada', 'minmod', 'no limiter']],
+            'smoother':[str, ['Runge-Kutta', 'DADI']],
+            'equationType': [str, ['Euler', 'laminar NS', 'RANS']],
+            'equationMode': [str, ['steady', 'unsteady', 'time spectral']],
```
nwu63, line 4520: time-spectral? Not a big deal either way

adflow/pyADflow.py
```diff
@@ -4516,259 +4482,244 @@ def setOption(self, name, value):
...
+            'NKUseEW':[bool, True],
+            'NKADPC':[bool, False],
+            'NKViscPC':[bool, False],
+            'NKASMOverlap':[int, 1],
+            'NKPCILUFill':[int, 2],
+            'NKJacobianLag':[int, 20],
+            'applyPCSubspaceSize':[int, 10],
+            'NKInnerPreconIts':[int, 1],
+            'NKOuterPreconIts':[int, 1],
+            'NKLS':[str, ['cubic', 'none', 'non monotone']],
```
nwu63, line 4621: similarly, non-monotone

doc/tutorial.rst
```diff
@@ -46,7 +46,7 @@ each statement line by line::
   'outputDirectory':outputDirectory,
 
   # Physics Parameters
-  'equationType':'RANS',
+  'equationType':'rans',
```
nwu63, line 49: These still need to be updated right?

tests/reg_tests/reg_default_options.py
```diff
@@ -30,12 +30,12 @@
     # Physics Paramters
     'discretization':'central plus scalar dissipation',
     'coarsediscretization':'central plus scalar dissipation',
-    'limiter':'vanalbeda',
-    'smoother':'runge kutta',
-    'equationtype':'euler',
+    'limiter':'van Albada',
```
nwu63, line 33: These options dictionary should be instantiated using `caseInsensitiveSet` to avoid duplicate entries (this was a bug I ran into a while ago which motivated this change)

nwu63 (commented):
A few minor comments

adflow/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '2.2.4'
+__version__ = '2.3.0'
```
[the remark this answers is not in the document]
sseraj: I agree we should wait for the other PR before making a new release. One of the two PRs will have the version bump though so I was going to keep this here. I can change this back if you prefer.

adflow/pyADflow.py
sseraj: No hyphen actually seems more common in the literature

anilyil:
> Yes, the users will have to update the string values. To communicate this, I am planning on listing the names of the affected string options in the release notes and linking to the options table for the new values. I can also add the list to the PR description.

Sounds good to me. @nwu63 do you request changes? I can approve this PR when things are settled.

adflow/__init__.py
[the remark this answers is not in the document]
nwu63: I think only the commit directly preceding the tag should contain the updated version. That way, if you check the version in Python you should be confident it is _at most_ as old as the tag, i.e. whatever change from the release notes have all been incorporated into the software.

adflow/pyADflow.py
sseraj: I think so. RK is still the best way to increase the residual after restart.

adflow/pyADflow.py
sseraj: I was actually thinking of making this one word based on how prevalent 'nonmonotone line search' is compared to 'non-monotone line search'.

nwu63: Haha I find `nonmonotone` incredibly illegible, but that's still better than `non monotone` IMO.

adflow/pyADflow.py
eirikurj: I agree. For string values we should make them more readable by separating using spaces when appropriate.

adflow/pyADflow.py
anilyil: `'infchangecorrection':True` is also a great option (better than RK reset imo). I think we are planning to make that enabled by default in #41

tests/reg_tests/reg_default_options.py
sseraj: I tried this and it breaks things elsewhere in the tests. I think we should leave this for a later PR.

nwu63: I'll open an issue for this.

nwu63 (approved):

adflow/pyADflow.py
```diff
@@ -4516,259 +4476,244 @@ def setOption(self, name, value):
...
-            'volumevariables':[list, ['resrho']],
-
-            # Multidisciplinary Coupling Parameters:
-            'forcesastractions':[bool, True],
+            'numberSolutions':[bool, True],
+            'printIterations':[bool, True],
+            'printTiming':[bool, True],
+            'setMonitor':[bool, True],
+            'printWarnings':[bool, True],
+            'monitorVariables':[list, ['cpu','resrho', 'resturb', 'cl', 'cd']],
```
anilyil, line 4672: Should we also update these options? It should be `resRho` etc. with the current convention right? I think these are more commonly used compared to other affected options, so I am also fine with not changing these.

nwu63: I suggest that we keep this for now. In the future, I plan to change the way these are set up in baseclasses. Like the list of choices, there should be a _set_ of all possible variables to monitor, and the user can specify a list (does the order matter? maybe for print out?) and we check in base classes if that is a subset of the total allowable set of monitor variables.

anilyil (approved):

PR #106, opened 2021-01-20
Azure transition

bernardopacini:
## Purpose
This is a draft PR for transitioning to Azure. The azure pipelines yaml file is included in the .github directory and the travis file is updated given the renamed input_files folder (from inputFiles).

## Type of change
What types of change is it?

- [ ] Documentation update
- [X] Maintenance update
- [ ] Other (please describe)

## Testing
All tests run with Azure pipelines.

## Checklist


.github/azure-pipelines.yaml
```diff
@@ -0,0 +1,31 @@
...
+extends:
+  template: azure/azure_template.yaml@azure_template
+  parameters:
+    REPO_NAME: adflow
+    IGNORE_STYLE: true
+    COMPLEX: true
+    GCC_CONFIG: config/defaults/config.LINUX_GFORTRAN.mk
+    INTEL_CONFIG: config/defaults/config.LINUX_INTEL.mk
+    INPUT_FILES: true
+    STEPS_MAKE_REAL: make
```
nwu63, line 26: The `STEPS_MAKE` and `STEPS_INSTALL` params are the default values right? So they don't necessarily need to be specified here.

.github/azure-pipelines.yaml
```diff
@@ -0,0 +1,31 @@
...
+    STEPS_INSTALL_REAL: pip install .
+    STEPS_TEST_REAL: testflo -v .
```
nwu63, line 28: Note that Travis has `-n 1` appended to the `testflo` commands, we probably want the same here.

.github/azure-pipelines.yaml
```diff
@@ -0,0 +1,31 @@
...
+- master
+
+resources:
+  repositories:
+  - repository: azure_template
+    type: github
+    name: mdolab/.github
+    endpoint: mdolab
+
+pool:
```
nwu63, line 14: Can this be specified in the template?

nwu63 (commented):
One more thing: I plan to update the Docker images so that there would be environment variables `$PETSC_ARCH_REAL` and $PETSC_ARCH_COMPLEX` which can be used instead of hard-coding it based on compiler etc. Because of that, I would suggest waiting for the new images to be built tomorrow, and then updating this PR accordingly.

.github/azure-pipelines.yaml
bernardopacini: This depends on how we want to handle repos that are pure python. If we want a conditional that will skip any steps to copy config files and make fortran if the make steps are empty, then we need to specify the lines in this file. Otherwise, we can use an additional boolean and have make defaults.

.github/azure-pipelines.yaml
bernardopacini: Got it -- I'll add this in.

.github/azure-pipelines.yaml
bernardopacini: Yes, it can be moved to be job-specific. I will add this change in the next push.

.github/azure-pipelines.yaml
nwu63: The approach I suggested was similar to 2, but instead of relying on `GCC_CONFIG` etc. to determine whether we have a build step, to check if `Makefile` exists. Similarly check `Makefile_CS` for whether complex build exists. This would be more general (since repos like TACS do not have these config files but still have make steps).

For these "weird" repos I also suggested having a bash script (stored alongside the yaml) for how to build the repo, instead of throwing the bash commands into this yaml file and cluttering it.

bernardopacini: Following Neil’s recommendation, I implemented the check for make files with the following lines:

```
  - script: |
      if { [[ ${{ parameters.BUILD_TYPE }} == 'real' ]] && [ -f "Makefile" ]; } || { [[ ${{ parameters.BUILD_TYPE }} == 'complex' ]] && [ -f "Makefile_CS" ]; }; then
        docker exec app /bin/bash -c ". \$HOME/.bashrc_mdolab && cd ${{ variables.DOCKER_WORKING_DIR }} && cp $(CONFIG_FILE) config/config.mk";
        docker exec app /bin/bash -c ". \$HOME/.bashrc_mdolab && cd ${{ variables.DOCKER_WORKING_DIR }} && ${{ parameters.STEPS_MAKE }}";
      fi
    displayName: Make

```

This is in the template (`.github` repo) so the change will take effect behind the scenes for this PR. `parameters.BUILD_TYPE` is hidden to the user and managed internally in the templates.

In terms of bash scripting make commands, I tested that already and it works fine, you just put the path of the bash script instead of the command in the respective variables. That being said, we may need to adjust the template if there is no config file to copy over.

nwu63: Is the `STEPS_MAKE` param set depending on `BUILD_TYPE`? For the config file, we can make it so that if it's an empty string, don't do the copying.

bernardopacini: No, build type is set as an additional constant that essentially just provides a flag. This ensures each test looks for the correct makefile:

```
- stage:
  dependsOn: []
  displayName: Test Real
  jobs:
  - template: azure_build.yaml
    parameters:
      REPO_NAME: ${{ parameters.REPO_NAME }}
      GCC_CONFIG: ${{ parameters.GCC_CONFIG }}
      INTEL_CONFIG: ${{ parameters.INTEL_CONFIG }}
      INPUT_FILES: ${{ parameters.INPUT_FILES }}
      STEPS_MAKE: ${{ parameters.STEPS_MAKE_REAL }}
      STEPS_INSTALL: ${{ parameters.STEPS_INSTALL_REAL }}
      STEPS_TEST: ${{ parameters.STEPS_TEST_REAL }}
      BUILD_TYPE: real

- stage:
  dependsOn: []
  displayName: Test Complex
  condition: eq(${{ parameters.COMPLEX }}, true)
  jobs:
  - template: azure_build.yaml
    parameters:
      REPO_NAME: ${{ parameters.REPO_NAME }}
      GCC_CONFIG: ${{ parameters.GCC_CONFIG }}
      INTEL_CONFIG: ${{ parameters.INTEL_CONFIG }}
      INPUT_FILES: ${{ parameters.INPUT_FILES }}
      STEPS_MAKE: ${{ parameters.STEPS_MAKE_COMPLEX }}
      STEPS_INSTALL: ${{ parameters.STEPS_INSTALL_COMPLEX }}
      STEPS_TEST: ${{ parameters.STEPS_TEST_COMPLEX }}
      BUILD_TYPE: complex
```

For the config file, we can implement that check as well.

nwu63: Okay that makes sense, I wanted to check that `STEPS_MAKE` was different dependent on whether it's a real or complex build.

.github/azure-pipelines.yaml
```diff
@@ -19,6 +19,7 @@ extends:
     COMPLEX: true
     GCC_CONFIG: config/defaults/config.LINUX_GFORTRAN.mk
     INTEL_CONFIG: config/defaults/config.LINUX_INTEL.mk
-    INPUT_FILES: true
-    STEPS_TEST_REAL: testflo -v . -n 1
-    STEPS_TEST_COMPLEX: export PETSC_ARCH=complex-opt-\$COMPILERS-\$PETSCVERSION && testflo -v . -m "cmplx_test*" -n 1
\ No newline at end of file
+    BUILD_REAL: .github/build_real.sh
```
nwu63, line 22: These can be set as default values in the template

.github/build_complex.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
nwu63, line 1: We may want `set -e` at the beginning (after the schebang) for these bash scripts

.github/test_complex.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+DIR=$(dirname $0)
```
nwu63, line 3: I think instead of this we should directly invoke `get-input-files.sh`, and same with `test_real`. This avoids code duplication.

nwu63 (changes requested):

.github/test_complex.sh
bernardopacini: Got it. I'll change that.

.github/build_complex.sh
bernardopacini: I'll add this in. Right now the `$CONFIG_FILE` variable isn't working, so those scripts are not final.

.github/azure-pipelines.yaml
bernardopacini: Yes, I did set the defaults in the template, I'll remove them from here.

nwu63:
Remaining TODO:
- [ ] Test that if you break the code then the build will catch it and fail as expected
- [ ] check if `testflo -n 1` is necessary, or if the default nprocs will actually be faster

sseraj:
The tests run slightly slower without ` -n 1`, so we should keep that in there.

sseraj:
As expected, all the tests fail after breaking the code.

sseraj:
@nwu63 @eirikurjThis should be good to go once the tests pass.

.github/azure-pipelines.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+extends:
+  template: azure/azure_template.yaml@azure_template
+  parameters:
+    REPO_NAME: adflow
+    IGNORE_STYLE: true
+    COMPLEX: true
+    GCC_CONFIG: config/defaults/config.LINUX_GFORTRAN.mk
+    INTEL_CONFIG: config/defaults/config.LINUX_INTEL.mk
+    PRIVATE: false
```
nwu63, line 22: Is this by default `true`? If not then this is not needed right?

.github/test_complex.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+./input_files/get-input-files.sh
+export PETSC_ARCH=complex-opt-$COMPILERS-$PETSCVERSION
```
nwu63, line 4: This line is actually not needed, since ADflow binary is linked to the correct PETSc library at compile time.

.github/azure-pipelines.yaml
sseraj: Good point. We are relying on the defaults wherever possible.

nwu63 (approved):

.github/test_complex.sh
sseraj: Maybe we still need this?

nwu63: Yep I guess I was wrong, my bad.

nwu63: Also, @sseraj would you mind adding the same to Travis? The tests on master are failing because of this erroneous commit I made a while ago to the master branch which removed the environment variable.

nwu63 (approved):

sseraj (approved):

Issue #99, opened 2020-12-12
Menter SST turbulence model outputs NaNs


DGCaprace:
## Description
Working on wind turbine blades with @marcomangano, I tried to use other turbulence models than the Spalart-Allmaras model which I had been using until now. I know that only `sa` is differentiated, but I just want to understand the sensitivity of the results to the turbulence model, from an analysis perspective. So I went for Menter's k-\omega SST model (see my options below), and I got NaNs and a PETSc error as off iteration 1 (see the log).

I also tried `k omega wilcox` and `k omega modified`, using `"turbresscale": [1.0e3,1.0e-6]`. There is no NaN in that case, however I found it suspicious that the two turbulence residuals displayed for monitoring are exactly equal to 0.0. Looks like it is the same at iteration 0 with Menter's model.

Less important: I noted a typo in the doc: the option to specify the rescaling parameters for turbulence models is `turbresscale` instead of `turbresscalar`.

Looking forward to your insights on what could be causing this.

## Options
```
{'adjointl2convergence': 1e-09,
 'adjointmaxiter': 500000,
 'adjointsubspacesize': 500,
 'adpc': True,
 'anklinresmax': 0.1,
 'ankmaxiter': 60,
 'anksecondordswitchtol': 0.01,
 'cfl': 1.5,
 'cflcoarse': 1.5,
 'equationtype': 'RANS',
 'gridfile': 'UAE_1blade_short_3pitch_coarse_5_L1_3D_128.cgns',
 'l2convergence': 1e-11,
 'l2convergencecoarse': 1e-05,
 'lowspeedpreconditioner': True,
 'mgcycle': 'sg',
 'monitorvariables': ['cpu', 'resrho', 'cd', 'resturb', 'yplus'],
 'ncycles': 30000,
 'ncyclescoarse': 250,
 'nkinnerpreconits': 2,
 'nkjacobianlag': 3,
 'nkouterpreconits': 3,
 'nksubspacesize': 100,
 'nkswitchtol': 1e-14,
 'nsubiterturb': 10,
 'restrictionrelaxation': 1.0,
 'smoother': 'dadi',
 'surfacevariables': ['cp', 'mach', 'yplus', 'sepsensor', 'p', 'temp'],
 'turbresscale': [1000.0, 1e-06],
 'turbulencemodel': 'menter sst',
 'useanksolver': True,
 'useblockettes': False,
 'usenksolver': True,
 'useqcr': True,
 'userotationsa': True,
[...]

PR #107, opened 2021-01-21
Bugfix on turbulence models

DGCaprace:
## Purpose
Adressing issue #99 on the use of turbulence models within the (A)NK solver:
* fixing NaN appearing when using `Menter SST`
* fixing turbulence residuals not computed in the (A)NK solver for all models except `SA`
See the issue for more details on the origin.

## Type of change

## Testing
I used the `tests/reg_tests/test_solve.py` script to test the fix on the tutorial wing using various turbulence models. Even though the code seems to be doing what it needs, I still need to check the flow solution more extensively.

Attached is the log of a run with Menter SST on the tutorial wing. The turbulence residuals are going up after Iter 141: I am not sure if this might still reflects a bug, or if this is 'normal'. Looks like SA also has a non-monotonous convergence in that case. 

[test_solve_ANK.log](https://github.com/mdolab/adflow/files/5851084/test_solve_ANK.log)

I tried to play with `turbresscale` but it did not change the order of magnitude of the residuals. I still need to check that.

## Checklist

- [ ] I have checked the quality of the flow solution and compared it across the various turbulence models available, and different meshes


DGCaprace:
The last commit is just me realizing that the code generates a runtime error during the regression tests when you compile it with GCC and `-fcheck=all`. The allocatable `qq` should in theory be allocated before entering any of the turbutils routine, even though it's not used in most of them.  Maybe you are already aware of that, and it's probably not critical. I am proposing a modification here but there are maybe constraints that I am not aware of when dealing with Tapenade. Anyway, we can revert that if it's not appropriate.

anilyil:
Thanks for the PR. I will take a look at it whenever I have time. Is this bug limiting your or @marcomangano's work in any way? I also requested reviews from other phd students who know a bit about adflow but no need for everyone to review. I just tagged you to make sure people know about this PR

DGCaprace:
Thank you. It's not really limiting our work, so nothing really urgent. I will continue testing this fix with other meshes, as part of the verification effort that we started on wind turbine geometries. Anyway, I look forward to you guys' comments on the code itself, whenever you have time.

DGCaprace:
I ran a bunch of simulations with the fix, essentially with Menter SST model, using a relatively coarse wind turbine mesh that @marcomangano and I trust.  Generally, it's much more difficult to obtain a converging behavior (w.r.t with the SA model), and I don't know if it is due to the model itself, or if there is still something wrong with the fix.

Here is what it usually does:  after a couple hundreds of smoothing iterations, the first ANK iterations goes ok but then the step size quickly goes down to 0.01, and stays like this forever, e.g.:

```
#---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
#  Grid  | Iter | Iter |  Iter  |   CFL   | Step | Lin  |    Wall    |        Res rho         |       Res kturb        |       Res wturb        |          C_Mx          |        totalRes        |         Y+_max         |
#  level |      | Tot  |  Type  |         |      | Res  | Clock (s)  |                        |                        |                        |                        |                        |                        |
#---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
      1     554    554     DADI   0.15E+01  1.00   ----  0.32875E+02   0.1298346565438176E+02   0.7232150264371758E-05   0.8780291662620172E+02  -0.6857612529621321E-01   0.8110883567607815E+04   0.2638563330528655E+01
      1     555    555     DADI   0.15E+01  1.00   ----  0.32934E+02   0.1234622743788891E+02   0.7439397567921133E-05   0.8559194625662319E+02  -0.7974965133049750E-01   0.7675113095667598E+04   0.2638864505781834E+01
      1     556    558     *ANK   0.10E+02  1.00  0.003  0.33417E+02   0.5001330186062396E+01   0.1849122553886089E-04   0.2004494259979415E+03  -0.9381478209566334E-01   0.2999525446231444E+04   0.2613624565383472E+01
      1     557    561     *ANK   0.42E+02  1.00  0.016  0.33823E+02   0.2056713133902179E+01   0.1653158769503024E-04   0.1758036928583075E+03  -0.4983211913437119E-01   0.1236648765940901E+04   0.2539150460052314E+01
      1     558    564     *ANK   0.16E+03  1.00  0.048  0.34224E+02   0.3541608938969110E+00   0.1177410795993622E-04   0.1386923906982015E+03  -0.1484892210408073E-01   0.4021591102434871E+03   0.2493339476561981E+01
      1     559    579    *SANK   0.76E+03  1.00  0.042  0.34803E+02   0.6722995131293009E-01   0.4375209484124058E-05   0.3991129628211742E+02  -0.2251377410039902E-01   0.9113013949052772E+02   0.2510411575499191E+01
      1     560    603    *SANK   0.45E+04  0.70  0.048  0.35512E+02   0.3174693987823816E-01   0.3098468267484977E-05   0.2774387057508451E+02  -0.2064034745785019E-01   0.7016807944433849E+02   0.2479496458332568E+01
      1     561    626     SANK   0.45E+04  1.00  0.046  0.35908E+02   0.1528129914763937E-01   0.1736077551313235E-05   0.1653454742576892E+02  -0.1784639897789380E-01   0.1812499156386824E+02   0.2458292229295459E+01
      1     562    701    *SANK   0.29E+05  0.00  0.116  0.37333E+02   0.1528131365274354E-01   0.1735891438824708E-05   0.1655118878146899E+02  -0.1784642378661237E-01   0.1818788415387252E+02   0.2458292198865495E+01
      1     563    776    *SANK   0.14E+05  0.00  0.088  0.38758E+02   0.1528131365274426E-01   0.1735891438824652E-05   0.1655118878144729E+02  -0.1784642378661240E-01   0.1818788415387175E+02   0.2458292198865494E+01
      1     564    851    *SANK   0.14E+05  0.01  0.088  0.40229E+02   0.1529606661973688E-01   0.2387058910124432E-06   0.1320281752868915E+02  -0.1736296757399657E-01   0.1801743669414549E+02   0.2459233099867608E+01
      1     565    926     SANK   0.14E+05  0.01  0.086  0.41394E+02   0.1530730439752916E-01   0.2578416971649194E-06   0.1403702329983747E+02  -0.1689671031760761E-01   0.1785152250274765E+02   0.2460137643965853E+01
      1     566   1001     SANK   0.14E+05  0.01  0.084  0.42558E+02   0.1531546051801310E-01   0.2716519944442218E-06   0.1439165889072453E+02  -0.1644448311928895E-01   0.1768388328179196E+02   0.2461012058947120E+01
``` 

Based on @marcomangano's suggestions and the Solver page of the documentation, I have tried the following:
- increasing `nsubiterturb`: no success
- using `ankunsteadylstol=1.5`: no success
- increasing `ankmaxiter`: helps to always meet LinRes<0.05, but doesn't improve step size or overall convergence
- reducing `ankcfllimit` to 1000: doesn't improve step size or overall convergence
- increase `ankpcilufill` to 4: doesn't improve step size or overall convergence
- use ANK for longer before switching to second order: slower convergence
- switching to `CANK`: leads the convergence to completely stall
- switching from `turbDADI` to `turbKSP`, with between 1 and 10 `anknsubiterturb`. This possibly helps to have larger step size, even though the CFL remains O(100); but the turb solver returns NaN/Inf after several iterations, from the DBG  info I get. This eventually cause the sim to crash.

Qualitatively though, the velocity field I also tried `k-w Wilcox`: it has a similar behavior.

At this point, I am happy to take any comment/suggestion that the above inspires you. I wonder if what happens with `turbKSP` could reflect a bug... and/or if the same type of diverging behavior of the turbulent solver can be the origin of small step sizes, even with `turbDADI`. Anyway, I will probably not spend much more time on this, unless the gurus have a clear insight of what else I could try. :-)

note: [Attached](https://github.com/mdolab/adflow/files/5884346/stdout_39691011.txt) the full output my baseline sim, with the full set of options I use.

anilyil:
Out of curiosity; why are you using `'ankswitchtol': 0.01,`? Do you need this option? If not, I would try w/o it and just do ANK alone instead of doing DADI. In general, DADI and ANK follow different convergence paths and I have been developing testing the ANK solver almost exclusively w/o using DADI before it. Switching to ANK at that point might cause the underlying continuation stuff to fail. 

The turbKSP returning NaN's is possibly an issue. During every Newton update, we have sth called "physicality check". This makes sure that the updates are physical. E.g. it only allows a 20% change in the density and energy terms at a time. This effectively makes sure that these values never get a negative value because they start with positive values. For the SA working variable, we allow it to decrease by 99% at each step but no limit on the other direction. This again ensures SA working variable will not be negative. We might need to do a similar physicality check for these turbulence models. Currently, the code just runs the same thing it does with SA, which may or may not be suitable for SST. Furthermore, I think the code only does this for one of the turbulence variables and skips the rest. Bottom line; turbKSP might need some implementation to get a robust solver with other turbulence models.

marcomangano:
If I remember correctly @DGCaprace reduced the `'ankswitchtol'` because he was having the same issues starting directly with ANK and thought that adding some smoothing could have helped convergence. So, I believe that changing that back to default will not help, but it might worth a shot to do similar tests with the options you reported in the previous comment, this time without the DADI smoothing.

DGCaprace:
That is correct. In this case, starting with the DADI smoother actually helps. Initially, I wasn't using it and figured out that the initial convergence of ANK is actually improved after smoothing. The fact that you would have expected the opposite makes me even more doubtful about the fix.  
Anyway, I reverted that and played a bit with the options above, and I could not get any better behavior. The baseline file I was proposing remains the best I have obtained so far. 

I will stop looking for now and wait for code review. Again, it would have been nice for our project without being a hard requirement in my opinion, so it's probably not urgent.

nwu63:
@anily what's the plan with this effort? Shelve this or do we want to investigate further?

anilyil:
@nwu63 we can keep this PR open and @DGCaprace can push updates as they find more stuff. I dont want to shelf it since there is some progress, but I dont want to merge it either because its not complete. Due to the non-time critical nature of it, I think its okay to let this PR sit. I think we can merge the master branch to here to prevent it from getting stale. I doubt we will make similar changes to the rest of the code that will cause conflicts here etc.

DGCaprace:
Alright I did some cleaning and reviewing on this PR, I think it is now ready for review by someone from the MDOlab. Let me know if there are things I can still improve.

**Summary of my edits in this branch:**
- Implementation of the [2003 version of the Menter SST model](https://turbmodels.larc.nasa.gov/sst.html). I got major improvements in convergence behaviors with this version.  Compared to the original, it mainly changes the way the eddy viscosity is computed. Both versions are still available in the code through a compilation flag, but it defaults to the 2003 version.
- `CorrectForK` in the pressure: it had not been carried over in the `computePressureSimple` routine. Menter SST does not converge without that.
- Apply clipping (physicality check) to all turbulence variables.
- Small fixes of wrong indices here and there.
- `computeEddyViscosity` for SST needs special care: it has to compute some derivatives of the velocity. So even when called with `includeHalos=false`, it will actually use the velocity in 1st halo cells. For ANK, it works fine. Problem is: in many other contexts, computeEddyViscosity is called when the 1st halo has not been exchanged yet… This is the case in rk4 (solver and smoother), DADI smoother, etc.  I have not tested those thoroughly but some improvements might be needed.
- I tested my latest edits with a NACA0012 airfoil at Re=5e6 and M=0.3. See Figs and run script attached. I used a mesh generated with pyHyp's example script.

CL_turbModels
CD_turbModels
[airfoil.py.txt](https://github.com/mdolab/adflow/files/8318747/airfoil.py.txt)

**Current limitations:**
- The 2-variable turbulence still won’t work with the coupled solver or TurbKSP. Obviously not with `(A)NKADPC` since we miss AD, but even with finite differences. I did not spend much time trying to fix that in the preconditioner. However, this branch at least enables analysis with ANK and SANK.
- No implementation for `blockettes=true`.
- `k-omega Wilcox` can still be subject to major convergence issues. 
- `k-tau` and `v-f` seem not to be working, even for the simple NACA0012. I could not identify exactly why.
- There are still a couple of minor questions I had: I marked them with a `QUESTION` tag in the code.

marcomangano:
As discussed with @anilyil , I redirected this PR to a newly-created, development branch in the mdolab repo. Unless there are some final quick fixes, I would recommend to merge this PR asap and continue the work on the bugfix at a later stage.

nwu63:
Whoever is merging this PR (I can also do this once @anilyil gives the go ahead) should _NOT_ squash. Instead, a simple merge should be fine, since we want to preserve these individual commits.

PR #111, opened 2021-01-25
Change to `printOptions` and fix docstrings

### LINKED ISSUE #108: Deprecation warnings
## Description
Running ADflow causes a number of DeprecationWarnings.
```
/home/mdolabuser/repos/adflow/adflow/pyADflow.py:3623: DeprecationWarning: invalid escape sequence \*
  """This the main python gateway for producing forward mode jacobian
/home/mdolabuser/repos/adflow/adflow/pyADflow.py:3800: DeprecationWarning: invalid escape sequence \*
  """This the main python gateway for producing reverse mode jacobian
/home/mdolabuser/repos/baseclasses/baseclasses/BaseSolver.py:184: DeprecationWarning: printCurrentOptions is deprecated. Use printOptions instead.
  warnings.warn("printCurrentOptions is deprecated. Use printOptions instead.", DeprecationWarning)
```
The first two are probably related to the way the docstrings are set up. The last one is due to an API change in `baseclasses`, simply rename `printCurrentOptions` with `printOptions`.

eirikurj:
## Purpose
Closes #108. 
This PR fixes the docstrings and now uses the `printOptions` functions.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -3651,7 +3651,7 @@ class ADFLOW(AeroSolver):
         Returns
         -------
         dwdot, funcsdot, fDot : array, dict, array
-            One or more of the these are return depending on the \\*Deriv flags
+            One or more of the these are return depending on the ``*Deriv`` flags
         """
 
         if xDvDot is None and xSDot is None and xVDot is None and wDot is None:
@@ -3833,7 +3833,7 @@ class ADFLOW(AeroSolver):
         Returns
         -------
         wbar, xvbar, xsbar, xdvbar, xdvaerobar : array, array, array, dict, dict
-            One or more of these are returned depending on the \\*Deriv flags provided.
+            One or more of these are returned depending on the ``*Deriv`` flags provided.
 
         """
         # Error Checking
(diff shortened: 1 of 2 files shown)


adflow/pyADflow.py
```diff
@@ -3651,7 +3651,7 @@ def computeJacobianVectorProductFwd(self, xDvDot=None, xSDot=None, xVDot=None, w
         Returns
         -------
         dwdot, funcsdot, fDot : array, dict, array
-            One or more of the these are return depending on the \*Deriv flags
+            One or more of the these are return depending on the \\*Deriv flags
```
nwu63, line 3654: Why do we need to escape the `*` character? Does it work without any slashes?

eirikurj: The `*` is an inline emphasis character which then starts but does not end. It will create an undesired behavior.

nwu63: Ah I see. In that case would it make more sense to surround it with backticks? i.e. ``*Deriv``

joanibal (approved):

adflow/pyADflow.py
eirikurj: Sure, that would work if we want this to be literal. Actually many other things in the docstring there should be done this way too. I will update this.

nwu63 (approved):

PR #112, opened 2021-01-27
Minor change to setOption

nwu63:
## Purpose
This is a minor change to setOption where casting to lower happens _after_ calling `super().setOption`. This ensures that the options set in baseclasses have the correct capitalization, e.g. for `printModifiedOptions`.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
All existing tests pass.

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -4322,9 +4322,8 @@ class ADFLOW(AeroSolver):
         """
         Set Solver Option Value
         """
-        name = name.lower()
-
         super().setOption(name, value)
+        name = name.lower()
 
         # If the option is only used in Python, we just return
         if name in self.pythonOptions:


sseraj (approved):

anilyil:
so for example, would this break the scripts that used `gridfile` instead of `gridFile`? I think I understand what the change is doing but I don't understand why you wanted to make this change?

eirikurj:
This change allows the options to retain the capitalization for example when printed out, either using `printOptions` or `printModifiedOptions`, through the `self.map` property of `CaseInsensitiveDict` (see https://github.com/mdolab/baseclasses/pull/38 for more details). Note that the option names in the `self.options` dict are still stored as *case-insensitive* here.

eirikurj (approved):

PR #116, opened 2021-02-04
Only print modified options when switching ap

nwu63:
## Purpose
Previously, the modified options were printed on every `__call__`, which is excessive. Now, they are printed only when switching aero problems, since it's possible for a solver to use different options based on aero problems.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
This works locally.

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -32,7 +32,6 @@ from petsc4py import PETSc
 from baseclasses import AeroSolver, AeroProblem, getPy3SafeString
 from baseclasses.utils import Error
 from . import MExt
-from pprint import pprint as pp
 import hashlib
 from collections import OrderedDict
 
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -915,10 +915,6 @@ class ADFLOW(AeroSolver):
         # Set filenames for possible forced write during solution
         self._setForcedFileNames()
 
-        # Remind the users of the modified options:
-        if self.getOption('printIterations'):
-            self.printModifiedOptions()
-
         # Possibly release adjoint memory if not already done
         # so. Don't remove this. If switching problems, this is done
         # in setAeroProblem, but for when not switching it must be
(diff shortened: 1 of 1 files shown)


sseraj (approved):

adflow/pyADflow.py
```diff
@@ -2546,10 +2542,12 @@ def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True):
         # Tell the user if we are switching aeroProblems
         if self.curAP != aeroProblem:
             newAP = True
-            if self.comm.rank == 0:
```
anilyil, line 2549: why did you remove the comm.rank check here? I am guessing the current version will print this on all procs in comm now?

anilyil (changes requested):
can you check and make sure the "switching to aero problem" print is not done on all procs. we just want the root doing it.

adflow/pyADflow.py
marcomangano: With the `baseclasses` refactoring, a [`pp` function that use pretty print on the sole root proc has been added](https://github.com/mdolab/baseclasses/blob/7ed195c5064b489325f74e4c6b551123b3b67a61/baseclasses/BaseSolver.py#L227-L238).
It is true however that the [pprint import](https://github.com/mdolab/adflow/blob/a310c0f8484b88dd74162de44f8f298d9be87b5b/adflow/pyADflow.py#L35) in `pyADflow.py` is ambiguous at this point, and we might want to remove it altogether and use the provided `baseclasses` method.
@anilyil @nwu63 @sseraj does it make sense?

anilyil: oh thats cool. yeah the import in pyadflow is misleading

nwu63:
I will fix the `pprint` import. Is the rest of the PR okay? I didn't change the rest of root print to use `self.pp` so maybe for consistency we want to revert back to the old code?

marcomangano:
> I didn't change the rest of root print to use `self.pp` so maybe for consistency we want to revert back to the old code?

We can update them when we reformat the python scripts imho

anilyil (approved):

sseraj (approved):
I agree with @marcomangano. I think this is fine as is.

PR #119, opened 2021-02-08
fixed max/min to match AD

joanibal:
## Purpose
The AD/CS of the max and min functions were not consistent when the real parts were equal. 
This was causing the derivatives of a flux routine to be inconsistent in the case I was testing. 
I changed the complexifed definitions of max/min to match tapenade. 

for example 

real code 
` max(dss(i,j,k,1), dss(i+1,j,k,1))`
 
FWD AD code 
```
if (dss(i, j, k, 1) .lt. dss(i+1, j, k, 1)) then
  y1d = dssd(i+1, j, k, 1)
  y1 = dss(i+1, j, k, 1)
else
  y1d = dssd(i, j, k, 1)
  y1 = dss(i, j, k, 1)
```
CS code 
` max(dss(i,j,k,1), dss(i+1,j,k,1))`
 
where max was defined 
```
complex*16 function max_cc(val1, val2)
  complex*16, intent(in) :: val1, val2
  if (real(val1) > real(val2)) then
    max_cc = val1
  else
    max_cc = val2
  endif
  return
end function max_cc
```
which equivalent to 
```
if (dss(i, j, k, 1) .gt. dss(i+1, j, k, 1)) then
  y1d = dssd(i, j, k, 1)
  y1 = dss(i, j, k, 1)
else
  y1d = dssd(i+1, j, k, 1)
  y1 = dss(i+1, j, k, 1)
```
which is different than the FWD AD code when dss(i,j,k,1) and dss(i+1,j,k,1) are equal. 

## Type of change
What types of change is it?

## Testing

## Checklist

- [NA] I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted


anilyil:
will take a look at these changes soon. one question; why do we have two complexify files?

joanibal:
I'm not sure. I believe the one in src is copied to src_cs anyway

anilyil:
then we should probably remove the one that is over-written?

src/modules/complexify.f90
```diff
@@ -437,60 +439,63 @@ function max_cccc(val1, val2, val3, val4)
...
     endif
-    if (real(max_cccc2) > real(max_cccc)) then
+    if ( real(max_cccc) < real(max_cccc2)) then
       max_cccc = max_cccc2
     endif
     return
   end function max_cccc
 
 ! MIN, intrinsic
+! the logical statements here are chossen to match fwd AD code from tapenade
```
sseraj, line 459: Should be 'chosen'

sseraj (commented):
The changes look good. I agree with Anil though that we should remove one of the files if we can.

anilyil:
@joanibal can you check why travis is failing? Do we need to update the references for the tests?

sseraj:
There was a bug in Travis. The tests running now should pass.

nwu63:
I deleted the following two files which were duplicate:
- `src/build/complexify.py`: I did not find a `cp` command for this script, so I assume this one is actually the duplicate and not the one in `src_cs/build/complexify.py`
- `src_cs/modules/complexify.f90`: this is not used since it's identical to the one in `src/modules/complexify.f90` and that one is definitely copied over during complexification.

There is a third duplicate file, `precision.f90` but I think this is actually supposed to be there since it's different for the two builds -- real vs. complex so I didn't touch that. Can someone go over this and make sure what I said is correct? I don't have a good understanding of the ADflow build process.

anilyil:
@joanibal the changes look good. I am running tests locally to make sure it all works (but tests also pass so it should be fine). One question: Sounds like we are trying to be consistent with tapenade. But which version? Can you check if newer versions of tapenade kept the same convention for min/max?

anilyil:
both real and complex tests passed for me locally, and I think the code makes sense. 

Regarding the copied complexify files; I think it makes more sense to keep the complex-only code in the src_cs directory. @nwu63 removed the complexify.py that was under src/build so that makes sense to me. However, the code currently copies the complexify.f90 in src/build to src_cs/build. The file is added here: https://github.com/mdolab/adflow/blob/master/src_cs/build/Makefile1#L17 I suggest we just track the complexify script under src_cs/build directory and not copy it.

Finally, the precision file under src_cs is slightly different than the real part. Also, it seems like it is not auto-processed like the other files so I am guessing keeping a copy under src_cs makes sense. See the line that removes it from the fortran file list, which is the list of fortran files that are auto-processed: https://github.com/mdolab/adflow/blob/master/src_cs/build/Makefile1#L21

anilyil:
Actually, to speed things up, I think this PR is good as is. There is a separate branch we are developing with @joanibal for the complex step modifications. We already modified a bit of the complex build process to make it faster etc, so I can make the relevant change for complexify.f90 there. This PR is good to go as far as I am concerned.

anilyil (approved):

nwu63 (approved):

PR #81, opened 2020-09-28
Fixed for rotating frame sensitivities, no history

anilyil:
## Purpose

This is a template PR we will use to review the changes required for the sensitivity fix with rotating reference frames. This PR contains a single commit with all the changes required, so the history is not there. This may not get merged eventually, but we will use this to review the code and see what parts we can clean up.

## Type of change
What types of change is it?
_Select the appropriate type(s) that describe this PR_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Explain the steps needed to test the new code to verify that it does indeed address the issue and produce the expected behavior.

## Checklist
_Put an `x` in the boxes that apply._

- [ ] I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #121, opened 2021-02-17
Bugfix for rotating frame sensitivities + add Tapenade checks

marcomangano:
## Purpose
Long overdue, this PR addresses the sensitivities inconsistencies on the current ADflow release when [setRotationRate()](https://github.com/mdolab/adflow/blob/b13a5424c624d2392aa55ff9abee9f8a87239fd5/adflow/pyADflow.py#L817) is included to add a rotation component to the wall BC. The source code fixes (up to [commit 0d39d74](https://github.com/mdolab/adflow/commit/0d39d74f6b1376a0632aae1399fbe62407682d2d)) have been initially implemented by @Aagaard87 to perform [high-fidelity aerodynamic shape optimization of a wind turbine](http://www.umich.edu/~mdolaboratory/pdf/Madsen2019a.pdf) using MACH-Aero.

The old dev-branch with the fixes has been rebased in the current ADflow release, and the code has been further cleaned for clarity. See the discussion on #81 for more details. The final code in this PR is mostly the same as in #81 , but the commit history has been fixed to preserve the development history.
An additional regression test has been added for the real build, while the complex build cannot be performed yet as convergence is obtained with the ANK solver, which is not complex-safe yet.

@SichengHe put a big debugging effort into restructuring the code to have the relevant subroutines relocated in the appropriate file, verify the consistency of the AD process, and ultimately debugging derivatives inconsistencies between adjoint and complex-step derivatives.

Note that this PR also introduces the Tapenade checks on ADflow - currently using version 3.10

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
A new test (real only) has been added to `test_adjoint.py`. The `.ref` file might need to be retrained on a docker image for consistency. Also note that the test will currently fail because the dedicated mesh has not been uploaded yet.

## Checklist


nwu63:
@marcomangano any idea why the rotating tests are failing?

marcomangano:
> @marcomangano any idea why the rotating tests are failing?

As I mentioned they use a different `.cgns` file... That I will send to you soon so it can be pulled down among other input_files :pray:

marcomangano:
The fix on the differentiated code will take more time than prevented, as it is not (only) related to Tapenade makefile.
I will most likely need to move around some key subroutines between different files to ensure consistency with the current code and eliminate the need for any post-differentiation edits.
THE CURRENT CODE IS NOT THE CODE WE ULTIMATELY WANT TO MERGE.
Will keep you posted, but I cannot provide a precise time estimate for this.

marcomangano:
Alright, as briefly described [here](https://github.com/marcomangano/adflow/pull/1) Sicheng fixed the misplaced routines that were causing a huge diff when running Tapenade. Now everything is consistent, although we had to unravel some elegant pointer loops.
In addition to the test I added, we also verified on a separate runcase that does an FD test. All looking good there.

The last step here is to check that all the remaining edits by Mads are consistent, then I will mark this as "Ready for review".
Heads up for the maintainers @anilyil @sseraj @joanibal bc there is a lot of beef here!

marcomangano:
OK, I apologize to everyone here, I am finally getting back to this. Before I go through the current code and check out the final code, there is a first "inconsitency" that @SichengHe found out: part of the diff we were getting after running Tapenade did not come from post-AD edits, but from bits of differentiated code that were not added before to the `_b` and `_d` files. Sicheng fixed these in the last few commits [1](https://github.com/mdolab/adflow/pull/121/commits/caa325f0fa886225403866fd4529a76d1f077afe), [2](https://github.com/mdolab/adflow/pull/121/commits/5dac0b6623557e26a1bdb93e619b075d6c907762), [3](https://github.com/mdolab/adflow/pull/121/commits/fd0bab65e28d32407252b74b96b55d6214241f3f), [4](https://github.com/mdolab/adflow/pull/121/commits/a0cdf383eb157c27984a557515958bd5ac220f2d), and [5](https://github.com/mdolab/adflow/pull/121/commits/b24b8439b80b161e09b366a0d2cf6fad5cdd00c4), but we would like to double check these.
@anilyil @Aagaard87 could you take a quick look there? I am pretty sure we discussed about this some time ago, was that just a few "uncommitted" changes on Mads' branch?

src/output/outputMod.F90
```diff
@@ -1393,6 +1393,7 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
     use utils, only : setPointers
     use flowUtils, only : computePtot
     use inputCostFunctions
+    use cgnsGrid, only : cgnsDoms,cgnsNDom ! see l. 4105 in preprocessingAPI.F90
```
marcomangano, line 1396: This should probably just refer to `subroutine updateRotationRate` in the mentioned file. I guess the use is self-explanatory, given the function name?

src/output/outputMod.F90
```diff
@@ -1499,7 +1507,25 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
        rangeFace(2,1:2) = rangeCell(3,1:2)
        iiMax = jl; jjMax = kl
 
-       ww1    => w(1,1:,1:,:);   ww2    => w(2,1:,1:,:)
+       ! mham have to get the mesh coordinates further down in order to compute
+       ! the correct Cp-normalisation for rotational setups with Wind Turbines.
+       ! Careful now; the flow variables, w(0:ib,0:jb,0:kb,1:nw), we point to
+       ! below uses what is defined in l.216-217, block.F90 as
```
marcomangano, line 1513: I guess we can refer to `type blockType` in `block.F90` here. Anything to add/remove about this comment block?

src/output/outputMod.F90
```diff
@@ -1782,17 +1823,94 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
...
+          ! ghostcell/halo. This explains: half*(pp1(i,j) + pp2(i,j))
+          ! From start of subr:
+          !       As the solution
+          !       must be stored in the center of the boundary face the average
+          !       value of the first internal cell and its corresponding halo is
+          !       computed. The counter nn is updated in this routine. However
+          !       it is not initialized, because multiple contributions may be
+          !       stored in buffer.
        enddo
+       ! mham: Comments on the computations above - no code
```
marcomangano, line 1878: This looks very good to have here imo.
(Gotta love a comment block that includes "as explained in the opening paragraph of this subroutine")

src/solver/solverUtils.F90
```diff
@@ -385,14 +385,14 @@ subroutine gridVelocitiesFineLevel(useOldCoor, t, sps)
        ! Set the pointers for this block.
 
        call setPointers(nn, groundLevel, sps)
-       call gridVelocitiesFineLevel_block(useOldCoor, t, sps)
+       call gridVelocitiesFineLevel_block(useOldCoor, t, sps, nn)
```
marcomangano, line 388: @SichengHe `nn` is a counter used for some loops that is now passed into the `_block` subroutines, right?

src/solver/solverUtils.F90
```diff
@@ -777,18 +782,30 @@ subroutine gridVelocitiesFineLevel_block(useOldCoor, t, sps)
...
-                        +         x(i-1,j,  k,  1) + x(i,j,  k,  1))
-                   xc(2) = eighth*(x(i-1,j-1,k-1,2) + x(i,j-1,k-1,2) &
-                        +         x(i-1,j,  k-1,2) + x(i,j,  k-1,2) &
-                        +         x(i-1,j-1,k,  2) + x(i,j-1,k,  2) &
-                        +         x(i-1,j,  k,  2) + x(i,j,  k,  2))
-                   xc(3) = eighth*(x(i-1,j-1,k-1,3) + x(i,j-1,k-1,3) &
-                        +         x(i-1,j,  k-1,3) + x(i,j,  k-1,3) &
-                        +         x(i-1,j-1,k,  3) + x(i,j-1,k,  3) &
-                        +         x(i-1,j,  k,  3) + x(i,j,  k,  3))
+                   xc(1) = eighth*(flowDoms(nn, groundLevel, sps)%x(i-1,j-1,k-1,1) &
```
marcomangano, line 785: This is the trick that Sicheng had to do for Tapenade to run - unravel the pointers such as x =>flowDoms(nn, groundLevel, sps)%x . @eirikurj @mdolab/adflow_maintainers any comments here? We could revert back to the more elegant code with pointers at a later stage, when Tapenade can handle this

src/solver/solverUtils.F90
```diff
@@ -835,114 +852,196 @@ subroutine gridVelocitiesFineLevel_block(useOldCoor, t, sps)
...
-                           + derivRotationMatrix(2,2)*xxc(2) &
-                           + derivRotationMatrix(2,3)*xxc(3)
-                      sc(3) = sc(3) + velzGrid           &
-                           + derivRotationMatrix(3,1)*xxc(1) &
-                           + derivRotationMatrix(3,2)*xxc(2) &
-                           + derivRotationMatrix(3,3)*xxc(3)
-
-                      ! Store the dot product of grid velocity sc and
-                      ! the normal ss in sFace.
+          ! i-direction
```
marcomangano, line 859: This is a bigger chunk of "unravelled" pointers loop, please take a look to see if this is good to merge

src/solver/solverUtils.F90
```diff
@@ -1193,7 +1295,9 @@ subroutine slipVelocitiesFineLevel_block(useOldCoor, t, sps)
           enddo
 
        enddo bocoLoop1
-
+#else
```
marcomangano, line 1298: do we usually add this `else continue` for `#ifndef`?

src/solver/solverUtils.F90
```diff
@@ -1348,60 +1430,387 @@ subroutine slipVelocitiesFineLevel_block(useOldCoor, t, sps)
...
-                xc(3) = fourth*(xFace(i+1,j+1,3) + xFace(i+1,j,3) &
-                     +         xFace(i,  j+1,3) + xFace(i,  j,3))
-
-                ! Determine the coordinates relative to the center
-                ! of rotation.
-
-                xxc(1) = xc(1) - rotCenter(1)
-                xxc(2) = xc(2) - rotCenter(2)
-                xxc(3) = xc(3) - rotCenter(3)
+          ! The new procedure is less elegant as the previous one.
```
marcomangano, line 1433: Another big loop to check, with a lot of conditionals

src/solver/solverUtils.F90
```diff
@@ -1499,48 +1908,136 @@ subroutine normalVelocities_block(sps)
...
+             else if (BCFaceID(mm) == iMax) then 
 
-                   ! Compute the normal velocity based on the outward
-                   ! pointing unit normal.
+               mult = one
 
-                   BCData(mm)%rFace(i,j) = weight*sFace(i,j)
-                enddo
-             enddo
+               do j=BCData(mm)%jcBeg, BCData(mm)%jcEnd
```
marcomangano, line 1939: And a last big loop

marcomangano (commented):
Alright, I went through the edits on the fortran code and highlighted the main edits. Please carefully check these - the code works fine but we want to be sure to commit to this approach before we proceed to merge.
I will check the python layer soon

adflow/pyADflow.py
```diff
@@ -569,8 +569,9 @@ def addSlices(self, direction, positions, sliceType="relative", groupName=None):
                 sliceName = "Slice_%4.4d %s Para Init %s=%7.3f" % (j, groupTag, direction, positions[i])
                 self.adflow.tecplotio.addparaslice(sliceName, tmp[i], dirVec, famList)
             else:
+                # e.g. Slice_0008 allWalls: Aboslute y=66.875
```
marcomangano, line 572: This should definitely be removed. Any reason why `self.adflow.tecplotio.addabsslice` was removed here? @Aagaard87

adflow/pyADflow.py
```diff
@@ -1448,6 +1449,235 @@ def evalFunctionsSens(self, aeroProblem, funcsSens, evalFuncs=None):
             print("| %-30s: %10.3f sec" % ("Complete Sensitivity Time", finalEvalSensTime - startEvalSensTime))
             print("+--------------------------------------------------+")
 
+    def evalFunctionsSensFwd(self, aeroProblem, funcsSens, evalFuncs=None):
```
marcomangano, line 1452: Is this function something we want to fix and maintain in the code for dot-product tests? 
I am 95% sure this is not working now, will check and add more info here.
Great comments though

adflow/pyADflow.py
```diff
@@ -3413,6 +3643,26 @@ def solveAdjoint(self, aeroProblem, objective, forcePoints=None, structAdjoint=N
                 self.curAP.adflowData.adjoints[objective] = psi
                 self.curAP.adjointFailed = False
 
+    def _process2_outer(self, dvName, TotalDeriv_):
```
marcomangano, line 3646: These are also only used in `evalFunctionsSensFwd()`

adflow/pyADflow.py
```diff
@@ -5631,6 +5881,42 @@ def _readPlot3DSurfFile(self, fileName, convertToTris=True):
 
         return pts, conn
 
+    # mham
+    def mhamDOT_slipvelocitiesfinelevel_block_d(self, useoldcoor, t, sps):
+        """ helper function to make dot product test"""
```
marcomangano, line 5886: This is never used in pyADflow but was most likely used during development to get the differentiated vals for slip velocities. The comments look good, but is it something we want to keep here or move to the docs?

marcomangano (commented):
Actually the additions in this layer were less convoluted than I expected. In the end it's a few functions added to get adjoints with fwd AD. Not currently working, do we want to try and fix them? If yes, do we want to address it in the current PR?

marcomangano:
FYI: the reason why the fwd sensitivities do not work might actually be related to `pyGeo`, see my error log (the regular sensitivity routine works fine):
```
Traceback (most recent call last):
  File "aero_opt.py", line 404, in <module>
    refSens = cruiseFuncsSens(x, refFuncs)
  File "aero_opt.py", line 219, in cruiseFuncsSens
    CFDSolver.evalFunctionsSensFwd(aeroProblems[i], funcsSens)  # for dot product test, not working
  File "/home/mmangano/miniconda3/envs/adflow-test/lib/python3.8/site-packages/adflow/pyADflow.py", line 1587, in evalFunctionsSensFwd
    residualDeriv_, funcDeriv_ = self.computeJacobianVectorProductFwd(
  File "/home/mmangano/miniconda3/envs/adflow-test/lib/python3.8/site-packages/adflow/pyADflow.py", line 4090, in computeJacobianVectorProductFwd
    xsdot += self.DVGeo.totalSensitivityProd(
  File "/home/mmangano/miniconda3/envs/adflow-test/lib/python3.8/site-packages/pygeo/DVGeometry.py", line 2015, in totalSensitivityProd
    dv = self.DV_listLocal[key]
KeyError: 'pitch'
```
Looks like pyGeo cannot find the DV  anywhere (in this case `pitch` is a global variable) when going for `totalSensitivityProd` calculation

marcomangano:
Alright, as I promised I am bugging you again on this, need your help to go through the comments and help me address the (hopefully) last fixes to this PR.

I also checked  `pyADflow.evalFunctionsSensFwd` more in detail, and it looks like it is mostly a wrapper for `computeJacobianVectorProductFwd` and for checking specific variables implementations, so I assume we can remove it altogether?

adflow/pyADflow.py
anilyil: So in general, I think there would be value in a "direct" implementation to get total derivatives. This however, is not that. It's still adjoint with a more inefficient way to get some of the partials using fwd AD. If you are not willing to implement direct properly, then I say you can remove this. However, there is still value in the explanations and the code etc to guide future implementations of direct. So I am not sure what to do here.

src/adjoint/Makefile_tapenade
```diff
@@ -228,6 +228,14 @@ oversetUtilities%newtonUpdate(xCen, blk, frac) > \
 oversetUtilities%fracToWeights(frac) > \
 			      (weights) \
 \
+solverutils%slipVelocitiesFineLevel_block(gammainf, pinf, timeref, rhoinf, veldirfreestream, machgrid, bcdata%uslip, flowDoms%x) > \
```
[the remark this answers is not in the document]
anilyil: does gammainf change? do we want to differentiate stuff with it?

src/adjoint/masterRoutines.F90
```diff
@@ -285,6 +286,9 @@ subroutine master_d(wdot, xdot, forcesDot, dwDot, famLists, funcValues, funcValu
     integer(kind=intType) :: ierr, nn, sps, mm,i,j,k, l, fSize, ii, jj, iRegion
     real(kind=realType), dimension(nSections) :: t
     real(kind=realType) :: dummyReal, dummyReald
+    ! mham
```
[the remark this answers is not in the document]
anilyil: I dont have a strong preference, either way is fine

src/adjoint/masterRoutines.F90
```diff
@@ -606,6 +636,10 @@ subroutine master_b(wbar, xbar, extraBar, forcesBar, dwBar, nState, famLists, &
     real(kind=realType), dimension(:), allocatable :: extraLocalBar, bcDataValuesdLocal
     real(kind=realType) :: dummyReal, dummyReald
     logical ::resetToRans
+    ! mham
+    real(kind=realType), dimension(nSections) :: t
+    logical :: useOldCoor ! for solverutils_d() functions
```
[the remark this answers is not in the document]
anilyil: I suggest renaming t since it is one letter so bad practice.

anilyil:
I see that a few `_fast_b` files were added in this PR. Do we need these? Do you use the differentiated routines here in the matrix free adjoint solution or do we only need them for partial evaluations? If we dont use them for the linear solver itself, we need to remove them from the PR.

src/output/outputMod.F90
anilyil: why do we need to access these from the `w` array? Can we get this info from BC face info?

src/solver/solverUtils.F90
anilyil: dont we use this pointer approach in the rest of the code? and tapanade works there right? is there a difference here that makes tapenade not work with our pointers?

src/solver/solverUtils.F90
anilyil: the precompiler commands are removed in the end when it is compiled. does that UseOldCoor if check work w/o the continue line or does it trip?

src/solver/solverUtils.F90
anilyil: can we include tests that check for every case (i,j,k/min,max)? this looks very error prone...

anilyil:
I went through the changes quickly and tried to comment on individual things I saw. We will need to have a call where you remind me all of these changes.

One high-level question: Do the new routines in master affect cases w/o a rotating reference frame? Do we run through the math regardless of rotation, or do you check and skip all of the code if there is no rotation? I am asking in terms of performance; there is no point in running this code for our regular stuff so we should make sure that when we use a stationary reference frame, the code does not do any unnecessary computations.

marcomangano:
@SichengHe can you go through Anil's comments? You are definitely more aware of the implementation/changes details

src/adjoint/masterRoutines.F90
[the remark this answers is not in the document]
marcomangano: @sseraj do you have strong opinions on this?

src/adjoint/Makefile_tapenade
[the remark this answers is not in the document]
marcomangano: @Aagaard87 @SichengHe I turn this question to you. I assume it was added for a good reason, but I need to check the code more in detail to figure out where are we using it

src/output/outputMod.F90
marcomangano: @SichengHe :eyes:

src/solver/solverUtils.F90
marcomangano: @eirikurj you had a good explanation for this if I remember correctly. I guess some other parts of the code are wrapped in a different way and so the pointers are not an issue. Further wrapping these subroutines was the alternative approach to this, @SichengHe opted for unravelling the loops as "safest" approach (right?)

src/solver/solverUtils.F90
marcomangano: (I am not really getting the question sorry, just bumping this up)

src/solver/solverUtils.F90
marcomangano: Are you talking about specific unit tests for this routine? I am onboard but don't really know where to start tbh

adflow/pyADflow.py
marcomangano: @sseraj also looking for your opinion here :pray: I could migrate the comments to a "dev guide" section of the docs?

adflow/pyADflow.py
sseraj: I would say this belongs in the docs

src/adjoint/masterRoutines.F90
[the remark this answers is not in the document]
sseraj: I think they should be removed for the most part

src/adjoint/masterRoutines.F90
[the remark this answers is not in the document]
sseraj: Agreed with renaming the variable

adflow/pyADflow.py
sseraj: I agree, it would be better in the docs.

src/adjoint/masterRoutines.F90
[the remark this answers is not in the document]
marcomangano: Suggestions for the name? :grin: I will fix this and cleanup the comments soon-ish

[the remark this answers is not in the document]
sseraj: I just realized it's called `t` because the same variable name is used in `gridvelocitiesfinelevel_block`, etc. Maybe just call it `time`?

marcomangano:
please @mdolab/adflow_maintainers check the last commits:
- Added back the "add absolute slices" function in pyAdflow, I am pretty sure it was erroneously deleted
- cleaned up the comments in the fortran layer, and moved all the good tips into the docs (pls take a look to ensure I wrote something correct and understandable)
- Removed the additional functions for the fwd sensitivities from `pyADflow`

SichengHe (commented):
@anilyil 
It is a good point. From what I have seen, I did not expect the update will change the performance a lot. The code I have modified is related to grid velocity, it is only called when the grid has been updated, and the residual computation will always take the grid velocity into consideration (see fluxes.f90). I am not very familiar with the part @Aagaard87 has modified.

SichengHe:
> I see that a few `_fast_b` files were added in this PR. Do we need these? Do you use the differentiated routines here in the matrix free adjoint solution or do we only need them for partial evaluations? If we dont use them for the linear solver itself, we need to remove them from the PR.

I think @anilyil made a good point here. I checked the MakeTapenade file, it seems that we did not add anything new to the state only derivative parts. So that means we shall not see new _fast_b file. I doubt that these files are manually generated. I checked several, and it seems that the file is not differentiated. We probably shall remove them if they are not called anywhere. Could you @marcomangano  check what happens if we remove them?

src/adjoint/Makefile_tapenade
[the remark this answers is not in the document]
SichengHe: I did not see why we differentiate gammaInf. gammaInf will change only when we consider different gas or temperature high enough...

src/solver/solverUtils.F90
```diff
@@ -555,11 +557,14 @@ subroutine gridVelocitiesFineLevel_block(useOldCoor, t, sps)
...
        end if
     endif
+#endif
 
     testMoving: if( blockIsMoving ) then
        ! Determine the situation we are having here.
 
        testUseOldCoor: if( useOldCoor ) then
+#ifndef USE_TAPENADE
+! We do not consider the finite difference based velocity computation for now
```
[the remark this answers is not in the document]
SichengHe: That is possible. That will be related to unsteady adjoint which is something far away...

src/solver/solverUtils.F90
SichengHe: Right. We need this because we need to directly access the data block in the function.

src/solver/solverUtils.F90
SichengHe: @anilyil I did not recall a function contains a locally defined pointer (something like x => y(3)) can be safely differentiated. If you have something in mind, I want to take a look.

anilyil: I am not sure what you mean about locally defined pointers. I think the differentiated code in adflow is full of pointers though. Consider this metric computation example which accesses the same x array as you have above:

Master routine calls `metric_block` (https://github.com/mdolab/adflow/blob/master/src/adjoint/masterRoutines.F90#L98) after calling `setPointers` (https://github.com/mdolab/adflow/blob/master/src/adjoint/masterRoutines.F90#L94). Then `metric_block` uses the pointers imported from `blockPointers` to run its calc using the array `x` https://github.com/mdolab/adflow/blob/master/src/adjoint/adjointExtra.F90#L185. Note it does not need to go through the block data structure, it just uses the block pointer.

Now lets look at how this is differentiated: `master_d` routine calls `setPointers_d` (https://github.com/mdolab/adflow/blob/master/src/adjoint/masterRoutines.F90#L373), which is a wrapper around the regular `setPointers` but also includes the correct pointers for the memory allocated to track the derivative seeds (https://github.com/mdolab/adflow/blob/master/src/utils/utils.F90#L3584-L3595). After `master_d` calls `setPointers_d`, it can then call `metric_block_d` routine (https://github.com/mdolab/adflow/blob/master/src/adjoint/masterRoutines.F90#L387), which is AD'd using the block pointers: https://github.com/mdolab/adflow/blob/master/src/adjoint/outputForward/adjointextra_d.f90#L400 

This way, the code still operates on block pointers and uses tapenade to differentiate against these block pointers. We manually handle the block pointers used for the derivative seeds by calling `setPointers_d`. I believe the AD pointers are also manually allocated by us in: https://github.com/mdolab/adflow/blob/master/src/modules/block.F90#L783-L785 I only used fwd AD examples here but same logic holds for reverse AD as well.

With all of this info in mind, I don't understand what is limiting the use of block pointers here. There already is a call to `setPointers` right before calling `gridVelocitiesFineLevel_block` in `solverUtils` in the current PR: https://github.com/mdolab/adflow/pull/121/files/db808880b542f45a484365090ce6c97f7e8373df#diff-c30a13783d79386e711f67eef5ddf3d84d06b73e8507b19b238d335e43dd1f0eL387 What we would need to do is to modify the `gridVelocitiesFineLevel_block` to operate on block pointers, and then manually differentiate `gridVelocitiesFineLevel` to have a call to `setPointers_d` before calling `gridVelocitiesFineLevel_block_d`. 

I may be totally missing something but this was my understanding of how ADflow is differentiated using whatever tapenade version w/ or w/o support for pointers.

src/solver/solverUtils.F90
anilyil: the `#ifdef` and `#endif` type statements are called precompiler directives I think. Those are the lines used to modify the actual code before it gets compiled. So if we are running a regular make w/o tapenade, and you want to exclude some lines, you do `#ifdef tapenade` or whatever the flag is, and because that flag will be false during a regular make, the precompiler will remove that chunk of the code. With that info, search a bit online to see how precompiler directives are actually used (my knowledge on this is tiny), and then these comments will make more sense. Then we can figure out whether we need them here or not.

anilyil:
I think the extra fast_b files are added because tapenade generates them anyways but we dont use them for compilation. A better approach would be to not generate the unused files, but I guess that would need modifications to tapenade makefile. In the past, I think people removed the extra fast_b files manually before committing. We can at least add them to the gitignore now so that in the future, they are auto-ignored instead of being tracked when someone adds all files in a commit

anilyil:
Re performance: I got it, so after volume coordinates change, we only run this calc once and not during residual evaluations right? And I am guessing there are flags to skip this computation if the grid is not moving. Can you verify that these routines are not executed when the grid is not rotating (or steady grid) @marcomangano @SichengHe ?

src/solver/solverUtils.F90
marcomangano: I am sorry, my reply was very general: I get the precompilers directives, and now I also understand which [if check](https://github.com/mdolab/adflow/blob/db808880b542f45a484365090ce6c97f7e8373df/src/solver/solverUtils.F90#L1146) you are referring to.. I just can't give you an answer, that's an old Mads' edit. Do you think is it worth to test this without the `else continue`?

Anyway, my question was more a general coding doubt, I think now I get the logic here, no need to change anything from my side.

marcomangano:
> I think the extra fast_b files are added because tapenade generates them anyways but we dont use them for compilation. A better approach would be to not generate the unused files, but I guess that would need modifications to tapenade makefile. In the past, I think people removed the extra fast_b files manually before committing. We can at least add them to the gitignore now so that in the future, they are auto-ignored instead of being tracked when someone adds all files in a commit

This makes sense, I confirm that `fast_b` was generated by Tapenade, and the manual patch explains why this was not already in the repo. I will just add these to `.gitignore` for now

marcomangano:
> Re performance: I got it, so after volume coordinates change, we only run this calc once and not during residual evaluations right? And I am guessing there are flags to skip this computation if the grid is not moving. Can you verify that these routines are not executed when the grid is not rotating (or steady grid) @marcomangano @SichengHe ?

I can check this with a bunch of print statements I assume, I can take care of it!

anilyil:
No need for print statements, you can trace the code and see how/when the routine is called.

src/solver/solverUtils.F90
marcomangano: I haven't started reading this yet but I assume this would fit in the docs :wink:

src/solver/solverUtils.F90
anilyil: ah I see, it was for an empty if then statement. yeah its ok to keep I guess

doc/adjoint.rst
```diff
@@ -95,5 +95,80 @@ The user can give a custom function that uses the intrinsic ADflow functions. AD
...
+We do *NOT* ``MPI_SUM`` the funcDeriv since that has already been done in the hand-differentiated code from ``surfaceIntegrations.F90``in the subroutine called ``getSolution_d()``.
+Now, we can finally compute the total sensitivity for every function of interest with:
+
+.. code-block:: python
+
+            TotalDeriv_ = funcDeriv_[myKey] + AllReduced_psi_dot_rDeriv
+
+Note that the final derivatives must be processed with ``_processAeroDerivatives()`` into a usable dictionary.
+
+Notes on dot-product tests
```
sseraj, line 139: The connection from the f2py steps to the dot product test is hard to grasp. I think the idea is to wrap both the forward and reverse mode AD routines with f2py and then call a dot product test function in Python. If that's correct, it would be best to make that more explicit and perhaps include what the Python code for the dot product would look like.

doc/adjoint.rst
```diff
@@ -95,5 +95,80 @@ The user can give a custom function that uses the intrinsic ADflow functions. AD
 
 The boundary condition definition and sensitivities are stored under ``AeroProblem`` to facilitate the use of different boundary conditions for multipoint cases.
 
-The reverse master (``master_b``) routine is wrapped by a thin wrapper which is the ``computeMatrixFreeProductBwd`` subroutine defined in ``adjointAPI.F90``. This wrapper allocates memory etc. before the derivatives are calculated. 
+The reverse master (``master_b``) routine is wrapped by a thin wrapper which is the ``computeMatrixFreeProductBwd`` subroutine defined in ``adjointAPI.F90``. This wrapper allocates memory etc. before the derivatives are calculated.
 
+ComputeJacobianVectorProductFwd
+-------------------------------
+
+This function is the 'twin' of ``ComputeJacobianVectorProductBwd`` and can be used to calculate the derivatves using the tangent AD code (i.e. forward code) instead of the usual reverse (adjoint) AD code.
```
sseraj, line 103: typo 'derivatves'

tests/reg_tests/test_adjoint.py
```diff
@@ -174,8 +174,68 @@ def span(val, geo):
     },
 ]
 
+# Adding this separately because we don't have the complex version available yet
```
sseraj, line 177: Are you able to add a complex test now that ANK is complex-safe? The test you added is good, but it doesn't prove that the bug has been fixed.

sseraj (commented):
I don't have much to add on the Fortran side, but I left some comments on the docs and tests

src/solver/solverUtils.F90
SichengHe: @anilyil  Yep, I got it that we can do AD for pointers. However, if you define y=>x where x is some array, then Tapenade struggles to differentiate y... That is what I mean by the locally defined pointer. I am not sure whether that is accurate nomenclature. @eirikurj may be able to explain this in more detail.

anilyil: Okay, I get what you mean by the locally defined pointer. Is this the problem here though? For example, do you need `flowdoms(...)%x` instead of just importing `x` from `blockPointers`?  Some other pointer stuff that might be in the code that I remember were related to faces, and those can be imported from BC data I believe, instead of a locally defined pointer to a slice of block data.

tests/reg_tests/test_adjoint.py
marcomangano: This is a good point, initially it was supposed to be added in a second moment... but probably it is better to not procrastinate anymore

doc/adjoint.rst
marcomangano: Yeah I need to double-check this section again bc it was mostly copy-pasted from comments

marcomangano:
I need to retrain the ref file as now we provide `mach` as an input - and the previously used `V` available as DV.
Hopefully the geom test will get to the end this time?

DavidAnderegg:
I am highly interested in this and would be willing to help a bit on the side (I see myself as some kind of support). Maybe i could generate a mesh or test something. I talked to @anilyil about this.

marcomangano:
Hi @DavidAnderegg, sorry for my late reply! Thanks for your offer to help. As you can see this PR has been dragging for a while but it is still not ready to be reviewed and merged - there are possible minor bugs that still need to be addressed.

We need to do some internal debugging in the next few weeks but as we get closer to "deployment" it could be helpful to have some extra checks from your side. I will bug you again when we get there! Thanks

DavidAnderegg:
Hi @marcomangano, this sounds great! Just let me know when I can help.

src/adjoint/Makefile_tapenade
```diff
@@ -174,22 +174,22 @@ sa%saResScale(scratch, dw) > \
...
-                             (w, pInfCOrr, rhoInf, p, fw) \
+fluxes%inviscidDissFluxScalar(w, pInfCorr, rhoInf, p, radi, radj, radk, fw, gammainf) > \
+                             (w, pInfCOrr, rhoInf, p, fw, gammainf) \
 \
-fluxes%inviscidDissFluxMatrix(w, pInfCorr, si, sj, sk, p, fw) > \
+fluxes%inviscidDissFluxMatrix(w, pInfCorr, si, sj, sk, p, fw, sfacei, sfacej, sfacek) > \
                              (w, pInfCorr, si, sj, sk, p, fw) \
 \
-fluxes%inviscidDissFluxScalarApprox(w, p, radi, radj, radk, fw) > \
+fluxes%inviscidDissFluxScalarApprox(w, p, radi, radj, radk, fw, gammainf, rhoinf, pinfcorr) > \
```
marcomangano, line 186: @anilyil @SichengHe I have just removed `gammainf`, do we need this added `rhoinf` and `pinfcorr` here?

marcomangano:
I just merged @SichengHe's fixes - the derivatives now have a very good match between the adjoint and the complex step.
The last issue to fix is the mismatch in the complex test I added here - probably due to some flow multimodality issue we observed elsewhere while debugging.
Once that is addressed, I will mark this PR ready for review!

nwu63:
Yep I'd be in favour of bumping the patch version.

marcomangano:
Tapenade is failing because the ADed TS code has not been added to ADflow yet. @SichengHe what are the plans for that? Shall we postpone the testing to that later PR? @nwu63

marcomangano:
(The tapenade check now works)

doc/adjoint.rst
sseraj: It would be good to clean up the docs, especially now that you have a lot more experience with this.

src/solver/BCRoutines.F90
```diff
@@ -1240,35 +1240,37 @@ subroutine bcEulerWall(nn, secondHalo, correctForK)
...
+          ww1(j,k,ivx)  = ww2(j,k,ivx) + vn*BCData(nn)%norm(j,k,1)
+          ww1(j,k,ivy)  = ww2(j,k,ivy) + vn*BCData(nn)%norm(j,k,2)
+          ww1(j,k,ivz)  = ww2(j,k,ivz) + vn*BCData(nn)%norm(j,k,3)
+
+          ! The laminar and eddy viscosity, if present.
+
+          if( viscous )    rlv1(j,k) = rlv2(j,k)
+          if( eddyModel ) rev1(j,k) = rev2(j,k)
+       end do
+    end do
```
sseraj, line 1272: Was this a bug in bcEulerWall?

tests/reg_tests/test_adjoint.py
```diff
@@ -326,6 +407,12 @@ def cmplx_test_geom_dvs(self):
 
         xRef = {"twist": [0.0] * 6, "span": [0.0], "shape": numpy.zeros(72, dtype="D")}
 
+        if self.name == "Rotating_wing":
+            rtol = 4e-6
+        else:
+            rtol = 1e-8
+        atol = 5e-10
```
sseraj, line 414: Are these the tightest possible tolerances for the non-rotating cases? We were using rtol of 5e-9 before, but that was with a different atol.

sseraj (commented):
Nice to see the match between the adjoint and complex step! I left a few comments below, including one from before on the documentation.

src/solver/residuals.F90
```diff
@@ -28,7 +28,10 @@ subroutine residual_block
     integer(kind=intType) :: i, j, k, l
     integer(kind=intType) :: iale, jale, kale, lale, male ! For loops of ALE
     real(kind=realType), parameter :: K1 = 1.05_realType
-    real(kind=realType), parameter :: K2 = 0.6_realType ! Random given number
+
+    ! The line below is only used for the low-speed preconditioner part of this routine
+    real(kind=realType), parameter :: K2 = 639.048910156_realType ! t2_=0.5 for wsp=8
```
sseraj, line 33: Also, I think we can remove the changes to the low-speed preconditioner. This line especially doesn't make any sense to me.

tests/reg_tests/test_adjoint.py
marcomangano: oh yeah my bad, I copy-pasted the tolerances from the previous test.

src/solver/residuals.F90
marcomangano: Yeah maybe @Aagaard87 can help here. I see how `K2` is used later on line 247 (but that variable `t2_` is not used anywhere, might be removed?) and on line 252 to calculate `betaMr2`.
There are a few comments that highlight doubts on the preconditioner implementation, and we know that it is not really working as intended.
@joanibal is planning to work on the low-speed preconditioner and might have some more insight.

src/NKSolver/blockette.F90
```diff
@@ -205,7 +213,21 @@ subroutine blocketteRes(useDissApprox, useViscApprox, useUpdateIntermed, useFlow
...
+                   time(mm) = time(mm) + (sps-1)*sections(mm)%timeperiod/real(&
+                         &         ntimeintervalsspectral, realtype)
+                end do
+             end if
+
+             call gridvelocitiesfinelevel_block(useoldcoor, time, sps, nn)
+             ! required for ts
+             call normalvelocities_block(sps)
+             ! required for ts
+             call slipvelocitiesfinelevel_block(useoldcoor, time, sps, nn)
```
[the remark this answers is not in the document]
marcomangano: @SichengHe can explain what happened, but looking at [this commit](https://github.com/marcomangano/adflow/commit/4671d1c477ade51ec934bfaa881ba1c62cd832e6) it looks like these changes were key to fix the discrepancies with the complex tests

marcomangano:
@sseraj agreed about the docs, especially since #174 has been merged. Maybe @joanibal and I can put a short paragraph together to explain how to run these tests - I haven't checked the new functions yet

SichengHe:
> Tapenade is failing because the ADed TS code has not been added to ADflow yet. @SichengHe what are the plans for that? Shall we postpone the testing to that later PR? @nwu63 Shall we temporarily exclude that part of the code from differentiation?

I think it makes sense to avoid differentiating that part for now. I am working on differentiating that part for now. Should be done soon.

src/solver/residuals.F90
SichengHe: I agree with @sseraj . I think probably we can remove all changes to low speed preconditioner thing for now since we are not really using it.

src/output/outputMod.F90
```diff
@@ -1504,7 +1511,25 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
        rangeFace(2,1:2) = rangeCell(3,1:2)
        iiMax = jl; jjMax = kl
 
-       ww1    => w(1,1:,1:,:);   ww2    => w(2,1:,1:,:)
+       ! have to get the mesh coordinates further down in order to compute
```
SichengHe, line 1514: i feel the documentation is a bit messy here. It is not a huge concern, but we may want to clean things up a bit... @marcomangano

SichengHe (changes requested):

.gitignore
```diff
@@ -24,6 +24,12 @@ src/build/libadflow-f2pywrappers2.f90
 src/build/libadflowmodule.c
 src/build/importTest.py
 
+# unused fast_b routines generated by Tapenade
```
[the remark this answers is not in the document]
marcomangano: @anilyil I thought I did something wrong but it looks like Tapenade is still generating `adjointextra_fast_b.f90`, `bcdata_fast_b.f90`, `oversetutilities_fast_b.f90`, and `zipperintegrations_fast_b.f90`. Shall we cleanup the Makefile or do they have any role in the differentiation?

src/solver/residuals.F90
marcomangano: @SichengHe @sseraj I am cleaning this up, but leaving most of the comments because there might be value in them... are you ok with that?

sseraj: That's fine with me.

src/output/outputMod.F90
marcomangano: @SichengHe let me know if it looks better now

SichengHe:
@anilyil For the solverUtils, we tried to use the pointers before but we ran into this issue (if I recall correctly): Tapenade did not know how to relate the pointers seeds in the subroutine to the underlying allocated pointer seeds in modules.  @eirikurj shall know better about this behavior. I am not sure why in some other cases the Tapenade behaves normally.

eirikurj:
I think the current solution is the best compromise. Its a little more verbose having to explicitly write all the directions, but the alternative is to rewrite the previous function and split into smaller parts so the pointers could be set outside. This would create a leaner code, but calls for a more significant refactoring and updates. However, I think the current implementation (were the loops are unrolled) could be improved by refactoring parts of the code into functions that each direction calls. This should reduce the duplication. I had a conversation with @SichengHe offline about this and discussed possible implementation.

marcomangano:
@eirikurj @anilyil @SichengHe please check https://github.com/marcomangano/adflow/pull/3 and let me know if anything else needs to be cleaned up or if you have better suggestions for the function name and where you want to have it.

SichengHe (approved):

src/adjoint/masterRoutines.F90
```diff
@@ -452,9 +472,13 @@ subroutine master_d(wdot, xdot, forcesDot, dwDot, famLists, funcValues, funcValu
...
 
+          call timeStep_block_d(.false.)
+
+          call initRes_block_d(1, nw, nn, sps)
+
           ! initalize the residuals for this block 
-          dw = zero
-          dwd = zero
+          ! dw = zero
+          ! dwd = zero
```
sseraj, line 481: These lines can be removed

tests/reg_tests/test_adjoint.py
```diff
@@ -172,6 +172,69 @@ def span(val, geo):
         "aero_prob": ap_tutorial_wing,
         "evalFuncs": ["fx", "mz", "cl", "cd", "cmz", "lift", "drag"],
     },
+    # # Rotating frame test
+    {
+        "name": "Rotating_wing",
+        "options": {
```
sseraj, line 178: I think some of these options can be removed, like the ones related to multigrid and NK. It doesn't look like either of those solvers are used. If we ever need to recreate the restart file, this would make it easier to see which non-default options were used.

tests/reg_tests/test_adjoint.py
```diff
@@ -287,6 +362,12 @@ def cmplx_test_aero_dvs(self):
             # this will happen when training, but will hopefully be fixed down the line
             return
 
+        if self.name == "Rotating_wing":
+            rtol = 2e-6
```
sseraj, line 366: Any chance these tolerances can be tightened with the fix in #203?

sseraj (commented):
I left some minor comments. These can be addressed later though if you think it will delay the PR. Otherwise, this looks good to go.

tests/reg_tests/test_adjoint.py
marcomangano: Unfortunately the tolerances need to be the same, when I tried smaller values I got the same relative error value as before

sseraj (approved):

src/adjoint/Makefile_tapenade
```diff
@@ -174,34 +174,37 @@ sa%saResScale(scratch, dw) > \
...
                                     (w, p, aa) \
 \
-flowUtils%allNodalGradients(w, vol, si, sj, sk, aa) > \
+flowUtils%allNodalGradients(w, vol, si, sj, sk, aa, ux, uy, uz, vx, vy, vz, wx, wy, wz, qx, qy, qz) > \
                            (w, vol, si, sj, sk, aa, ux, uy, uz, vx, vy, vz, wx, wy, wz, qx, qy, qz) \
 \
-fluxes%viscousFlux(w, x, si, sj, sk, rlv, rev, aa, ux, uy, uz, vx, vy, vz, wx, wy, wz, qx, qy, qz, fw) > \
-                  (w, x, si, sj, sk, fw, viscsubface%tau) \
+fluxes%viscousFlux(w, x, si, sj, sk, rlv, rev, aa, ux, uy, uz, vx, vy, vz, wx, wy, wz, qx, qy, qz, fw, viscsubface%tau, viscSubface%q) > \
+                  (w, x, si, sj, sk, rlv, rev, aa, ux, uy, uz, vx, vy, vz, wx, wy, wz, qx, qy, qz, fw, viscsubface%tau, viscSubface%q) \
```
anilyil, line 205: why do we need all of these additions here?

src/adjoint/Makefile_tapenade
```diff
@@ -156,14 +156,14 @@ turbBCRoutines%applyAllTurbBCThisBlock(rev, w, bvtj1, bvtj2, bvtk1, bvtk2, bvti1
 turbBCRoutines%bcTurbTreatment(w, rlv, d2wall, winf) > \
                               (w, rlv, d2wall, winf, bvtj1, bvtj2, bvtk1, bvtk2, bvti1, bvti2) \
 \
-solverUtils%timeStep_block(w, pInfCorr, rhoInf, si, sj, sk, p) > \
```
anilyil, line 159: pressure is removed here, any specific reason or was it not used anyways?

anilyil (commented):
I went over the code and left a few comments. I think most should be cosmetic, or me not understanding stuff and asking for an explanation. Besides my individual comments, one thing I am not getting is why we are now including the variable `dtl` in the AD codes. I did not leave comments everywhere this showed up to keep it cleaner. It looks like it has to do with how the gridvelocities stuff is called in master routines, but I would love to have a more complete explanation of the reason here.

Other than my review, I think we are also waiting for @eirikurj to review this as well? So I will just leave these as comments.

src/output/outputMod.F90
```diff
@@ -1823,17 +1858,93 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
...
     case (cgnsCp)
-
        ! Factor multiplying p-pInf
-
        fact = two/(gammaInf*pInf*MachCoef*MachCoef)
 
        do j=rangeFace(2,1), rangeFace(2,2)
           do i=rangeFace(1,1), rangeFace(1,2)
              nn = nn + 1
-             buffer(nn) = fact*(half*(pp1(i,j) + pp2(i,j)) - pInf)
```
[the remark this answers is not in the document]
marcomangano: check_cp
Yeah, good point. I am sure @Aagaard87 has been very diligent with his derivation, but I double checked to be safe. The Cp curves above are from the MACH-Aero tutorial wing, red is from this PR and green the current `main`. Perfect match at all stations

src/output/outputMod.F90
```diff
@@ -1823,17 +1858,93 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
        !================================================================
 
     case (cgnsCp)
-
        ! Factor multiplying p-pInf
-
        fact = two/(gammaInf*pInf*MachCoef*MachCoef)
```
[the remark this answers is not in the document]
marcomangano: good catch, I am removing it

src/solver/solverUtils.F90
marcomangano: There is no way to track the coverage in Fortran, but `gridVelocitiesFineLevel_block` is called by `gridVelocitiesFineLevel` and that subroutine is used by both the rotating frame and the unsteady cases, so if anything is broken here we should have major failures or errors in the tests we added.
Any opinions on this @SichengHe ?

src/adjoint/Makefile_tapenade
marcomangano: This is just the diff being weird, p is still there, right after `sFaceK`

src/adjoint/Makefile_tapenade
marcomangano: @SichengHe prefers to include all inputs/outputs to avoid partials to be zeroed out by mistake as it was happening before. Maybe not the most efficient approach, but the safest. Should we be more selective here?

marcomangano:
@anilyil looks like the diff on `residuals.F90` will remain odd because we moved `initres_block`out of the `#ifndef USE_TAPENADE` section of the code

src/adjoint/Makefile_tapenade
SichengHe: @marcomangano That is correct. Not 100% sure about the behavior of adflow, maybe @eirikurj has more comments on this.

src/output/outputMod.F90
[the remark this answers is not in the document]
anilyil: I checked the formulations and both are equivalent for _most_ use cases. However, moving away from using MachCoef breaks some functionality where we can prescribe it separately. This is used for a few cases, sounds like including TS: https://github.com/mdolab/adflow/blob/main/adflow/pyADflow.py#L3015-L3025 

To avoid breaking this functionality, I suggest we modify this part. We can compute the original factor's components outside the loop using machcoef and then in the loop we can keep including the rotational term corrections for the denominator. 

I can probably make this change myself, or give you instructions on how to fix it. Whichever you prefer @marcomangano

src/adjoint/Makefile_tapenade
anilyil: I dont know the details here but it sounds like being more selective would be better in terms of development effort later on. It's similar to the use statements that add everything; it becomes difficult to understand what terms are relevant and what terms are not.

anilyil:
It looks like my only remaining comments are on including all variables in the tapenade makefile and the cp output computation. The PR is fine for me other than those.

src/adjoint/Makefile_tapenade
marcomangano: Ok I gave a quick look at the code here, and it looks like all these terms are related to viscous flux calculations. I think only `flowUtils%allNodalGradients` and `fluxes%viscousFlux` have extra terms, all the other additions look targeted.
Figuring out which of these terms should not be differentiated goes a bit beyond my current understanding of the code, and I don't think that iteratively removing components until the gradient  breaks again is a sound approach.

@SichengHe @anilyil do you have a bit more insight/suggestions here?

marcomangano:
> Alright @anilyil @SichengHe @eirikurj , I tried to cleanup the tapenade makefile by removing some of the dependent variables that "did not look dependent" (i.e. I eyeballed), I tested the adjoint locally but I missed all the jacobian vector product failures happening right now. As @nwu63 mentioned, there's no point in removing stuff by trial and error because even if all the tests pass, we might be missing some zeroed-out partials.
> 
> My hot take is that ADflow maintainability in this sense has been already compromised, and we should make a separate effort to fix that and 1) cleanup `Makefile_tapenade` and 2) update it to work with Tapenade 3.16.

I stand by this, but I managed to fix and complete the cleanup of these routines. All other additions look targeted for the bugfix!

src/output/outputMod.F90
```diff
@@ -1823,17 +1858,93 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
...
+       !
+       ! Necessities                       units
+       ! U_i - inflow speed                      [m/s]
+       ! w   - rotational rate                   [rad/sec]
+       ! L_  - the span at point (i,j,k)         [m]
+       ! P_i - farfield pressure                 [N/m^2]
+       ! P_0 - pressure at surface point (i,j,k) [N/m^2]
+       !
+       ! fomula for apparent velocity:
+       ! (U_a)^2 = ([U_i,0,0]^T - [0,0,(2*pi*L_)/(2*pi)[m/rad]*w[rad/sec]])^2
```
sseraj, line 1940: How are the units of `(2*pi*L_)/(2*pi)` `[m/rad]`? It seems like it should just be `[m]`. Why even bother with this derivation? The only part relevant for the code is the last line `(U_a)^2 = (U_i)^2 + (L_*w)^2`, and even that doesn't explain the cross-product between `L_` and `w`.

sseraj (commented):
Nice work on the AD. However, I read through the commented code in outputMod.F90 in earnest for the first time, and I have several comments. You don't have to resolve them all, but some clean up would be nice.

marcomangano:
> Nice work on the AD. However, I read through the commented code in outputMod.F90 in earnest for the first time, and I have several comments. You don't have to resolve them all, but some clean up would be nice.

Good call, I had that in mind but then forgot to actually fix it. Check it out now and let me know if the comments make sense

sseraj (approved):
Much better, thanks!

sseraj (dismissed):

anilyil:
I want to do another quick pass some time later this week but I think the PR is good to go. Why were we not getting failures due to the extra `timestep_d` call? Was it because the contributions there are small or we dont test for that?

marcomangano:
> I want to do another quick pass some time later this week but I think the PR is good to go. Why were we not getting failures due to the extra `timestep_d` call? Was it because the contributions there are small or we dont test for that?

I think it was because of the small contributions, there are no unit tests but we are definitely covering that part of the code with the rotating test we added

sseraj (approved):
This is good to merge for me. Do we want to wait for @SichengHe's review?

anilyil:
I am fine with merging it. I just hit re-request on everyone who was involved. I think @marcomangano can do the honors

sseraj:
Go for it @marcomangano and please make the release afterwards as well

Issue #55, opened 2020-06-10
Update ANK to work with complex step


anilyil:
## Type of issue

- New feature (non-breaking change which adds functionality)

## Description
Currently ANK solver does not work with the complex step (CS) method. Furthermore, the behavior of the NK solver with (CS) is not well documented. We should fix and document both solvers with complex step so that they can be used for CS verification studies.

marcomangano:
I can work on this, maybe together with @bernardopacini and @nwu63

PR #128, opened 2021-03-04
Updated default options for testing

### LINKED ISSUE #104: Update default options list in tests
## Description
The default options list specified [here](https://github.com/mdolab/adflow/blob/836b203b5222a9e252d9ccd7fc964a00cd0e3145/tests/reg_tests/reg_default_options.py#L12) should be reviewed and updated
- we had decided to use a minimal list of options that mirror common usage, rather than specifying everything. That way, if the default options have changed, this will detect it. Also, this is closer to common usage (which also relies on default options)
- this dictionary should use `caseInsensitiveDict` such that each test which calls `update()` can be sure that no duplicate keys with different capitalization gets entered.

sseraj:
## Purpose

I updated the default test options to be as close to the actual defaults as possible without reducing test coverage or breaking tests. The main change is that the ANK solver is now tested. The converged solutions are the same as without ANK so I did not have to change any reference values.

Two exceptions for this:
1) ANK is not used for the adjoint tests because it does not work in complex mode yet. Once #55 is resolved, ANK should be used for the adjoint tests. 
2) The time-accurate test fails marginally with ANK. I turned ANK off for this test.

An added bonus of using ANK is that the tests should run slightly faster.

Closes #104

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
Very good! Actually, in the test I added for the rotating adjoint, I added a (sketchy?) [workaround](https://github.com/marcomangano/adflow/blob/1145dbbac3e28ee6778e1148e8e9b6ef5cf52236/tests/reg_tests/test_adjoint.py#L168-L229) to use ANK in the real (and only) test. You could use a different set of options (you reduced it to a few lines of code anyway) for real and complex tests, if you don't want to wait for the complex ANK to be merged.

Otherwise, you can just move all the tests to ANK and then overwrite the switch tol within [TestCmplxStep](https://github.com/mdolab/adflow/blob/b8cb700697c44190ad1f35666a050575343638b9/tests/reg_tests/test_adjoint.py#L227)?

joanibal (approved):

joanibal:
This seems good to me. 

@marcomangano. 
Right not, the solution is loaded using a restart value for the real adjoint tests (the ones that actually use the adjoint).

Using ANK is only going to speed up the solve tests for now and the complex adjoint tests once complex ANK is a working.

marcomangano:
> This seems good to me.
> 
> @marcomangano.
> Right not, the solution is loaded using a restart value for the real adjoint tests (the ones that actually use the adjoint).
> 
> Using ANK is only going to speed up the solve tests for now and the complex adjoint tests once complex ANK is a working.

Yeah that was a dumb idea now that I think about

nwu63 (approved):

PR #129, opened 2021-03-06
added an optional limit on the overset flooding iterations

anilyil:
## Purpose
This PR adds the option to specify an iteration limit to the flooding algorithm.
Normally, the flooding algorithm iterates until the flooded cells stop changing between the last 2 iterations.
However, this can be problematic when the first overset iteration causes all of the mesh to flood.
To debug these meshes, users have been instructed to modify the fortran routine previously.

With this PR, the option 'nFloodIter' can be set in the python level to limit the number of flooding iterations.
The default value is -1, which preserves the original behavior of the algorithm where it runs until the flooded cells stop changing.
The `nFloodIter` option can now be set to 1 (or 2) along with `nRefine = 1`, and then the output volume file can be used to understand how the mesh is flooding.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Old overset behavior is preserved with the default option.

## Checklist


sseraj (dismissed):
Nice feature. Thanks for adding a detailed description in the docs as well.

sseraj (approved):

joanibal (approved):

PR #130, opened 2021-03-09
Complex ANK+NK

anilyil:
## Purpose

This PR improves the complex ANK and NK solvers, as well as improving how complex residuals and printouts are handled. With @joanibal, we tried to figure out how the complex step method worked for implicit solvers. We concluded that the path of convergence is not important, but in the end, complex residuals must be converged to obtain accurate derivatives. This means that during a complex-step evaluation, we dont need to re-set flow, and just call the NK/ANK solvers repeatedly so that complex residuals converge. This PR has a few changes that makes this possible.

Here is a list of what we did with this PR:

1. The residual printout is fixed in complex mode. The code used to print the square of the residual, and the imaginary part of the printed value is equivalent to an interesting derivative (derivative of the solver update wrt complex-perturbed DV), but it was not printing the true complex residual. Because we need to converge the complex residuals, we print these directly now. (interestingly, complex residuals go to zero also when the derivative of the solver update wrt complex-perturbed DV also goes to zero, but the residual norm is easier to understand and is always positive)
2. The funcional printout is also improved. The output now looks like this:
```
#-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
#  Grid  | Iter | Iter |  Iter  |   CFL   | Step | Lin  |    Wall    |            Res rho              |                    C_drag                       |            totalRes             |
#  level |      | Tot  |  Type  |         |      | Res  | Clock (s)  |                                 |                                                 |                                 |
#-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
      1       0      0     None     ----    ----   ----  0.39389E-02   0.14551813E-14 +0.11287700E-129i  0.4710284646449747E-01 +0.1156509033894532E-123i  0.45644342E-12 +0.84458180E-128i
      1       1     21      *NK     ----    1.00  0.000  0.70712E+00   0.13697775E-14 +0.15267370E-132i  0.4710284646449758E-01 +0.1156509263873306E-123i  0.33574505E-12 +0.77222624E-131i
```
We print all 16 digits of real and complex parts for the functionals because the real part is the value and the complex part is the derivative.  However, we do not print all digits for residuals since their magnitude is more important anyways. I have left about 8 digits on both real and complex parts of the residuals for debugging purposes.  The nonlinear iteration header is also fixed so that things are aligned. I also explicitly give 3 digits to the exponent because we are now interested in doing smaller complex-steps that underflow and do not affect real parts when the real values are zero.

3. The complex build process is slightly improved. Complexify.f90 is moved to src_cs. We now recompile only the changed files in complex mode. We used to do this for fortran modules but still would copy over other include files which caused long build times. This is fixed. 

4. The absolute tolerance (`atol`) in the NK solver is set separately for complex mode. For the real mode, we set this atol to be 0.01 lower than the real L2 convergence target. This makes sure that the NK solver solves the linear system well enough to put us across the finish line, while not doing any extra work. This approach does not work with complex step. Because derivatives lag, complex residuals must be converged, even after the real ones hit machine zero. Furthermore, if we want to converge the complex system without re-setting the flow, then the real residuals will hang around machine zero and we again need the linear solver to actually solve the system and not quit early. Same fix is also applied to the ANK solver.

5. ANK solver is modified to handle complex mode. This involved several changes to the iteration algorithm as well as the line searches. I am still not 100% on some details, but it works, and the parts I dont understand go to zero as the CFL number is increased, so it should be fine. We only really want to avoid having the complex part diverge. As long as a relatively stable and low value of complex state is maintained, the complex system needs to be converged after the real part converges anyways due to the lag, so again, it should be fine.

6. We added a nonlinear solver test for complex ANK/NK (but see the unfinished item no 2 below). 

Here is a list that we did not do yet, and should be done in a future PR:
1. The complex build process needs to be cleaned up a bit more. This is not terribly bad, I just did not want to delay this PR for that because the complex build does work now. This will be a future PR.
2. We should add better ANK and NK solver tests. There are a few features that we want to maintain, and the simple test we added here can be modified to include more tests. Depending on feedback, we can do this in this PR, or in a future PR. 
3. The complex residual computation squares the real and complex parts separately to compute the two separate residuals. This will underflow with a complex step of 1e-200 and the residuals will just print zero. We ideally want to fix this square=>squareroot process so that the values do not underflow. 

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
We added complex-step tests for the ANK and NK solvers. 

## Checklist


anilyil:
So I still have a few TODOs before this is ready to merge. The fixes only work for decoupled ANK. I will need to fix the coupled ANK and turbANK until the PR is ready. Also, I want to add a test for multiple modes of ANK/NK in both real and complex modes. Finally, I think we can also update the existing complex tests to use ANK/NK with this PR. 

I wanted to create the PR to track these TODOs, and also have azure test the changes. Any feedback is welcome. 

@nwu63 I plan on fixing the complex compile in a separate PR. I have most of it laid out mentally, just need to figure out a few details. Once I create that PR, someone else can apply the similar changes to idwarp (I believe its the only other code we have with a similar complex build process?)

anilyil:
The tests fail because of the complex ank/nk test. That wasnt tagged as `cmplx_` as it should be. I will take care of it soon. We will need to update input files as well

src/NKSolver/NKSolvers.F90
```diff
@@ -1658,9 +1669,9 @@ module ANKSolver
   real(kind=realType) :: ANK_CFLMin0, ANK_CFLMin, ANK_CFLMinBase, ANK_CFLExponent
   real(kind=realType) :: ANK_stepMin, ANK_StepFactor, ANK_constCFLStep
   real(kind=realType) :: ANK_secondOrdSwitchTol, ANK_coupledSwitchTol
-  real(kind=realType) :: ANK_physLSTol, ANK_unstdyLSTol
+  real(kind=realtype) :: ANK_physLSTol, ANK_unstdyLSTol
```
sseraj, line 1672: Just taking a quick look at this PR. Is there a reason for this change? (And similarly two lines below)

anilyil: no, `realType` to `realtype` changes are just leftovers from me copy pasting. I think I tried setting them to `alwaysRealType` at some point but reverted back to `realType`. I can fix them back to the usual capitalization

anilyil:
This one is good to go for the review. I have added solver combination tests that switch between smoother, ANK, SANK, CSANK, and NK. I test all equation types with these, also with the 2 turbulence solvers that can be used in ANK. I still test the turbulence solvers with Euler and laminar NS because in the past, we had a few issues with these cases seg-faulting. With the current test, we just expect equation modes that does not have any turbulence model to just work regardless of the turbulence options specified. I test both real and complex versions with these. 

The test itself uses a simple 2^3 block originally made by @joanibal. I have modified the mesh slightly so there is a wall, and rest of the BCs are farfield. The flow also comes in at an angle to the wall so that Euler residuals are not zero with free stream. 

I have also modified existing complex tests to use ANK for the solution. @sseraj can you please check if I did it correctly? For the euler tests at least, I had to set the second order switch to get it to converge; it would oscillate a bit otherwise. Now it looks it is converging okay, but I dont really have a reference on how fast it should be. 

The only remaining item that I wanted to possibly do with this PR is the "unfinished" item 3 I listed above. A complex step size of around 1e-160 and below will underflow with the residual computation and the residual monitor will print zero for complex residuals. We can maybe include that in the "complex magic" PR that does a few fancy things. We can also update the step sizes in complex tests in that PR so that they use the underflow version. I will create an issue about this so we remember what is going on.

nwu63:
> The only remaining item that I wanted to possibly do with this PR is the "unfinished" item 3 I listed above. A complex step size of around 1e-160 and below will underflow with the residual norm computation and the residual monitor will print zero for complex residuals. We can maybe include that in the "complex magic" PR that does a few fancy things. We can also update the step sizes in complex tests in that PR so that they use the underflow version. I will create an issue about this so we remember what is going on.

Just to keep things moving, I think I am fine with this in a separate PR as long as it's documented and you remember to do it in the future (either as a separate PR or together with the upcoming complex PR).

anilyil:
Sure, I will create the issue. We can make a new release with this, or just do one release after also merging the actuator zone stuff in PR #118. That one was pretty much ready to go with tests and everything, but I needed complex ANK to run the complex tests in there. How about we make a new release once that is also merged @nwu63 ?

anilyil:
closes #55

tests/reg_tests/test_solver_combos.py
```diff
@@ -0,0 +1,217 @@
...
+
+# common options dict for both real and complex
+gridFile = os.path.join(baseDir, '../../input_files/cube.cgns')
+commonTestOptions = {
+    'gridfile': gridFile,
+    "writevolumesolution": False,
+    'writesurfacesolution': False,
+    'monitorvariables': ['cpu', 'resrho', 'resturb', 'cd'],
+    "mgcycle": "SG",
+    "mgstartlevel": -1,
```
sseraj, line 96: Probably don't need this option

tests/reg_tests/test_adjoint.py
```diff
@@ -75,12 +75,11 @@ def span(val, geo):
             "l2convergence": 1e-14,
             "monitorvariables": ["cpu", "resrho", "totalr"],
             "adjointl2convergence": 1e-14,
-            "mgcycle": "2w",
+            "mgcycle": "sg",
```
sseraj, line 78: Do we need to start with ANK for all the complex tests? I know that you added the solver combos test, but it would be nice to still have DADI/RK as part of a complex regression test (not just convergence).

sseraj (commented):
I'm still reviewing the code changes, but I have a few comments on the tests:

1) I also had some trouble with the complex Euler tests in #128. Somehow an `NKSwitchTol` of 1e-5 did not work but 2.5e-4 did. I don't think speed is an issue because the complex tests are running quite a bit faster with ANK.

2) Why use a 2^3 cube? Would it be useful to have a 3^3 cube or larger instead so that there is at least one internal cell?

tests/reg_tests/test_adjoint.py
anilyil: thats fair, I guess I wanted to see if everything would still work that way. To test it even more, I can go back to smoother+MG and then switch to ank and then to nk. That way tests touch more solvers basically. Does that make sense @sseraj ?

anilyil:
Regarding point 1: One of the euler tests failed to converge with 2 procs but converged with 1 proc. Thats why I probably added second order switch. I am not sure why NK acted like that when you worked on it, we can check the output and figure out why it failed that way. These are small cases that run on small number of procs so interestingly robustness goes down a bit since all of our stuff is tuned for larger problems.

Pt 2: As far as I know, once you have 2 layers in each cell its the same code that runs as if you had more layers (vs if you had 1 layer in one direction, code goes into "pseudo-2D mode"). I dont have a strong preference over it, I think @joanibal uses meshes like these to develop and test his research stuff. If you think a 3^3 would be better, we can change the mesh @sseraj.

Finally, thank you for reviewing the PR. This is one of those PRs where I would really appreciate the detailed feedback. Once  you leave all your comments, I will push the changes for the tests so that they use multigrid again.

sseraj:
1) Right, I probably should have checked the output to see what was going on. The laminar NS and RANS tests also have non-default values for `NKSwitchTol` so I was fine with adding one for Euler as well.

2) If 2^3 is general enough, then no need to use a larger mesh.

tests/reg_tests/test_adjoint.py
sseraj: Yes, that would be great.

src/NKSolver/NKSolvers.F90
```diff
@@ -3043,15 +3061,15 @@ subroutine physicalityCheckANK(lambdaP)
     real(kind=realType), pointer :: wvec_pointer(:)
     real(kind=realType), pointer :: dvec_pointer(:)
     real(kind=alwaysRealType) :: lambdaL ! L is for local
-    real(kind=realType) :: ratio
-
+    real(kind=alwaysRealType) :: tmp ! to receive the global step
```
sseraj, line 3064: I would prefer a name like `lambdaP_tmp` so it's easier to keep track of what's happening.

sseraj (commented):
Overall, the code changes look good from what I understand. Some clarification questions and comments below

anilyil:
Thanks for the review @sseraj, I will go through your comments and make the suggested changes. @joanibal can you review this anytime soon? I would rather address your comments together with addressing @sseraj's comments.

src/NKSolver/NKSolvers.F90
```diff
@@ -3166,13 +3220,24 @@ subroutine physicalityCheckANK(lambdaP)
...
+    ! so we will use the tmp variable to receive
+    call mpi_allreduce(lambdaL, tmp, 1_intType, MPI_DOUBLE, &
          mpi_min, ADflow_comm_world, ierr)
     call EChk(ierr,__FILE__,__LINE__)
 
+#ifndef USE_COMPLEX
+    lambdaP = tmp
+#else
+    ! finally, as a safety check, purge the complex part of lambda
+    lambdaP = cmplx(tmp, 0.0_realType)
```
joanibal, line 3238: So lambdaP is always real?
lambdaL is always real

joanibal: Shouldn't all lambda's always be real

joanibal: if all lambdas are always real, then you could probably get rid of your `tmp` variables

src/solver/solvers.F90
```diff
@@ -1613,12 +1624,24 @@ subroutine convergenceInfo
...
+
              if (monNames(mm) == cgnsL2resRho) then
                 rhoRes = monGlob(mm)
              end if
           case ('totalR')
+#ifndef USE_COMPLEX
              monGlob(mm) = sqrt(monGlob(mm))
+#else
+             ! take the square roots separately in complex mode
+             monGlob(mm) = cmplx(sqrt(real(monGlob(mm))), sqrt(aimag(monGlob(mm))))
```
joanibal, line 1643: We should consider adding a separate monitor for the norm of dR/dx.

As i mentioned on the issue you made for printing complex residuals. I don't think it is a good idea to compute h*|dR/dx|. It is best to first calculate the derivative and then take the norm of it.

tests/reg_tests/test_solver_combos.py
```diff
@@ -0,0 +1,217 @@
...
+# represent the most common use cases. Testing turbulence solver options with euler
+# and laminar NS makes sure the solver does not segfault etc.
+
+solver_combo_params = [
+    # Euler Tests
+    {
+        "name": "euler_smoother_ank_sank_csank_nk_turbdadi",
+        "options": {
+            "equationtype": "Euler",
+            # euler fails to get to machine zero in parallel...
```
joanibal, line 36: Why is that?

joanibal (changes requested):

joanibal:
Added a few suggestions in my review

joanibal:
So, to be clear. 

Does this resolve the issue of the slow convergence of the imaginary part after the real part has converged?

anilyil:
So slow is a bit relative term here and I am not sure what you have in mind.

After this PR, the complex parts still lag a bit. But, even the small amount of lag is okay for matching 10 digits of accuracy. So for most practical cases, we can just converge the real part tightly (1e-15) and the complex convergence you get there is good enough to get a 10 digit match with the adjoint. You can do more iterations to further nail down the complex residuals, and the accuracy increases (I mean difference between CS and adjoint results get smaller as expected). 

The changes to the `atol` in NK type solvers also mean that even when the complex residuals lag and the real part converged, you can just do one more call to the solver and that makes significant progress in the complex part. This is because before this PR, the linear solver would quit early because it would hit the atol limit. Now because atol is so much lower, it does more work solving the linear system even when real part is converged.

I have also tested this w/o resetting the real flow. With these cases, the real residual starts from machine zero whereas complex residuals start on the order of the complex step. Here, the complex convergence depends on the linear solver tolerance. With a linear solver tolerance of 1e-6, complex residuals converge after like 2-3 calls to the CFD solver. This also supports the theory we have been developing.

Does this answer your last question @joanibal?

src/NKSolver/NKSolvers.F90
```diff
@@ -589,7 +589,17 @@ subroutine NKStep(firstCall)
     ! convergnce check can't do anything either. By multiplying by
     ! 0.5, we make sure that the linear solver actually has to do
     ! *something* and not just kick out immediately.
+
+#ifndef USE_COMPLEX
+    ! in the real mode, we dont want to "oversolve the linear system" as detailed above
+    ! so we set the atol slightly lower than the target L2 convergence
     atol = totalR0*L2Conv*0.01_realType
+#else
+    ! in complex mode, we want to tightly solve the linear system every tine
```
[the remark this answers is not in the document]
anilyil: done.

joanibal:
Yes it does. 

Thank you

src/NKSolver/NKSolvers.F90
```diff
@@ -589,7 +589,17 @@ subroutine NKStep(firstCall)
     ! convergnce check can't do anything either. By multiplying by
     ! 0.5, we make sure that the linear solver actually has to do
```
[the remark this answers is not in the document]
anilyil: Let me clarify; The linear residual is of similar order as the nonlinear residual because the linear system's RHS is actually equal to the nonlinear system's residuals. Even though they are not identical (the linear system may have an initial guess, the way you compute the norms can be different, etc etc) they have a similar magnitude.

In here, the term "oversolving" refers to a different kind of oversolving prevented by the EW. The EW algorithm makes sure that you are not solving the linear system too tightly when you are far from the solution. The idea is that you are better of taking a more approximate step rather than taking a very accurate step, which will not give you too much nonlinear convergence to begin with. Here, what we mean by oversolving is this: Assume you are at a relative nonlinear convergence of 1e-11, and your target nonlinear convergence is 1e-12. You go into your Newton solver and now you need to solve a linear system. EW algorithm knows that you are close to the solution, so it will give you a relative convergence of a very low value, say 1e-6. Now, because you were so close to the L2 convergence target, you most likely could just solve the linear system by 0.1, and that will probably give you the nonlinear convergence you need. So this is what we mean by avoiding the oversolve. If you dont have an atol here, you will need to solve the linear system to 1e-6, whereas you could just do a bit of work and quit early because you are already there in the L2 convergence.

Now, there is another pitfall here. Lets go back to the previous numbers, you are at 1e-11 L2 convergence and target is 1e-12. If you set the atol value as `totalR0*L2Conv`, then the linear solver will do about 0.1 convergence and then quit. But nonlinear convergence can only be as good as your linear convergence, and often it is slightly worse than your linear convergence because the linear convergence is a linear approximation of your nonlinear system. So if you converge your linear system to 0.1, your nonlinear system may converge by 0.11. Now think what the solver would do; The top level nonlinear solver thinks there is still progress to be done since you are not exactly at 1e-12 (you were at 1e-11 and only converged 0.11 more), and the linear sytem thinks you are below the absolute tolerance for linear convergence because of you you just defined this atol. So they just go back and forth until you stop the code (it will run out of total iterations and quit with a "failed" solve). To avoid this issue, we multiply the atol by 0.01, so that we actually guarantee to solve past the nonlinear target, but only go past this target by 2 orders of magnitude (instead like 6 or 8). 

FINALLY, the complex system changes things a bit. Remember, we are trying to avoid oversolving the linear system, and we determine this by the residual. When running in complex mode, the complex residuals will lag, and we are interested in maybe doing more iterations to the solver to converge the complex residual, even though the real residual is already converged. However, if we keep the same atol of 0.01 lower than the L2 target, we can only solve the linear system by 0.01! Instead, I would much rather doing a tight solve (I can manually set the newton linear convergence to 1e-6) and nail down the complex residual, so that is why I have a different atol in the complex mode. This problem is caused by the complex residuals being negligible compared to real residuals, even when the real residuals are at machine zero.

This turned into a lecture, but whatever. Does this help @sseraj ? I wont spell check this so let me know if there is stuff thats not clear still. I would appreciate if you can suggest some changes to the comment to clarify this for people new to these ideas, or maybe we can link this comment in the code?

src/NKSolver/NKSolvers.F90
```diff
@@ -3032,7 +3050,7 @@ subroutine physicalityCheckANK(lambdaP)
     use flowVarRefState, only : nw, nwf, nt1, nt2
     use inputtimespectral, only : nTimeIntervalsSpectral
     use utils, only : setPointers, EChk, myisnan
-    use communication, only : ADflow_comm_world
+    use communication, only : ADflow_comm_world, myid
```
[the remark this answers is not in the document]
anilyil: it is not, and I was using it to debug back in the day. Removing it.

src/NKSolver/NKSolvers.F90
```diff
@@ -3166,13 +3220,24 @@ subroutine physicalityCheckANK(lambdaP)
     call EChk(ierr,__FILE__,__LINE__)
 
     ! Make sure that we did not get any NaN's in the process
-    if (isnan(lambdaL)) lambdaL = zero
+    if (isnan(real(lambdaL))) then
```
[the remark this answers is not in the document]
anilyil: good catch. I think this is left like this because I had `lambdaL` as a `realType` previously. Now that its `alwaysRealType`, we dont need this check. Removing it

src/NKSolver/NKSolvers.F90
```diff
@@ -3372,7 +3457,15 @@ subroutine ANKTurbSolveKSP
         rtolLast = rtol
 
         ! Set all tolerances for linear solve:
+#ifndef USE_COMPLEX
+        ! in the real mode, we dont want to "oversolve the linear system" as detailed above
```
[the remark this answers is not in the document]
anilyil: yes. I clarified by adding that we explain this in NKStep

src_cs/build/Makefile1
```diff
@@ -28,29 +26,39 @@ complexify:
 		echo Checking $$file $$realFile $$cmplxFile; \
 		if [ $$realFile -nt $$cmplxFile -o ! -e $$cmplxFile ];  then \
 			python complexify.py $$realFile; \
+			echo Copying $$file $$realFile $$cmplxFile; \
```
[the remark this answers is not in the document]
anilyil: done

src/NKSolver/NKSolvers.F90
anilyil: changed all instances of tmp in the two physicality checks to `lambdaP_recv`

src/NKSolver/NKSolvers.F90
```diff
@@ -3075,25 +3094,41 @@ subroutine physicalityCheckANK(lambdaP)
...
                 do j=2, jl
                    do i=2, il
-                      ! multiply the ratios by 10 to check if the change in a
-                      ! variable is greater than 10% of the variable itself.
+                      ! multiply the ratios by ANK_physLSTol to check if the change in a
+                      ! variable is greater than ANK_physLSTol of the variable itself.
 
                       ! check density
+#ifndef USE_COMPLEX
+                      ! to have the real mode sliiiightly more efficient, just do stuff with real numbers
```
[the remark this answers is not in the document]
anilyil: Yeah, so in the real mode, I dont need to check if the variable is real. I will rephrase the comment

src/NKSolver/NKSolvers.F90
```diff
@@ -3075,25 +3094,41 @@ subroutine physicalityCheckANK(lambdaP)
...
 
                       ! increment by 4 because we want to skip momentum variables
                       ii = ii + 4
 
                       ! check energy
+#ifndef USE_COMPLEX
+                      ! see the comment above for the difference between real and complex versions
                       ratio = abs(wvec_pointer(ii)/(dvec_pointer(ii)+eps))*ANK_physLSTol
+#else
+                      ratio = abs(real(wvec_pointer(ii))/real(dvec_pointer(ii)+eps))* real(ANK_physLSTol)
```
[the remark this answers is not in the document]
anilyil: So I was just trying to failsafe things. I realized that I could not set fortran-level options from pyADflow in the complex mode using the `alwaysRealType` for the data type. As a result, even though I would have preferred to have these variables be always real, I have to use `realType` for them. Locally, the complex part is set to zero by default, but I am not sure if all compilers would give this result so I just check. Do you have a better suggestion here @joanibal?

src/NKSolver/NKSolvers.F90
anilyil: I needed the `tmp` because the MPI_MIN operation in the MPI_Allreduce is not defined for complex numbers. It just fails. So I have to use an always real type to do the allreduce.

anilyil: I actually had a comment right above as well:
```
! mpi allreduce is not defined for complex numbers with the min operation
! so we will use the lambdaP_recv variable to receive
```

src/solver/solvers.F90
anilyil: You just cannot square the complex residuals when they are that small, and need to scale them. You and I understand what derivative that ends up being, but for a general public, its just so much easier to have a simple L2 norm defined that they need to converge. For now, I am happy with what the code does and it works reliably for all of our current practices. I will leave that modification to you when you move the code to use 1e-200 by default.

anilyil:
I addressed the comments in the fortran code. Now I will go over the tests again. I left all of the conversations as unresolved, please make sure that I addressed your comments @joanibal and @sseraj

tests/reg_tests/test_solver_combos.py
anilyil: I dont know. The linear solver result changed enough to trip the solver. Euler cases are more challenging for us to converge now because they dont have as much dissipation as the RANS ones. With that particular tuning, the Newton just stalled about that convergence level. I can probably just play around and make it converge more but it will probably require changing the linear solver tolerance in Newton manually etc. Do you want me to make this case converge or are you okay with the current state @joanibal ?

tests/reg_tests/test_solver_combos.py
anilyil: were you referring to the mgstartlevel? If so, removed it. By the way, does it matter if the mgcycle's value is `'SG'` or `'sg'`?

sseraj: Yes, I was referring to `mgstartlevel`. 

I realized recently that `'SG'` vs. `'sg'` doesn't matter because the string is passed directly to Fortran and Fortran is case insensitive.

anilyil: neat, I am switching to 'sg' to be consistent. removed the mgstartlevel

anilyil:
@sseraj is there a particular reason why we have 1e-16 convergence on the rans complex cases? It just hits machine zero and bounces around. Tests pass with an L2 convergence of 1e-15 as well, so I am thinking of just modifying that. Any objections here?

sseraj:
> @sseraj is there a particular reason why we have 1e-16 convergence on the rans complex cases? It just hits machine zero and bounces around. Tests pass with an L2 convergence of 1e-15 as well, so I am thinking of just modifying that. Any objections here?

No particular reason that I can think of. If all the tests pass with 1e-15, then that should be fine.

anilyil:
With my recent changes to the tests, I believe I addressed every comment from both of you. Please make sure if all is good. If the tests also pass, this is good to go for me.

src/NKSolver/NKSolvers.F90
anilyil: oh I just realized, `lambdaP` is not `alwaysReal`, it is just `realType`. As a result, I needed that additional always real variable to receive the MPI allreduce's output

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
sseraj: Thanks for writing this up! This helps a lot. I would suggest 3 changes to the comments to clear things up:
1) Add that the linear and nonlinear residuals have similar magnitude _at the start of the linear solution_, assuming a small initial guess
2) Mention that KSPSolve exits when either the rtol or atol is satisfied (not both), which means that the atol only comes into play near the end of the nonlinear solution
3) The block comment in NKStep does not actually mention oversolving the system so I would change your comment to just say "in the real mode, we set the atol slightly lower than the target L2 convergence as detailed above"

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
sseraj: The comment in `ANKStep` still needs to be changed

src/NKSolver/NKSolvers.F90
sseraj: This works for me

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
anilyil: Done. Please take a look at the new version

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
anilyil: changed the comments in `ANKStep` and `ANKTurbSolveKSP` (which does the same thing for the turbulence system)

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
sseraj: The new comments look great. Thanks! There are some typos left over from before but nothing major.

src/solver/solvers.F90
joanibal: > You just cannot square the complex residuals when they are that small,

Yep I got that 

> For now, I am happy with what the code does and it works reliably for all of our current practices. I will leave that modification to you when you move the code to use 1e-200 by default.

sounds good

tests/reg_tests/test_solver_combos.py
joanibal: > Do you want me to make this case converge or are you okay with the current state @joanibal ?

This seems important to verify the derivatives to the highest possible accuracy. 
With that being said, I don't want to hold this up. Maybe it's best to merge and add an issue.

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
joanibal: >  I realized that I could not set fortran-level options from pyADflow in the complex mode using the alwaysRealType for the data type

I have done this. I could be misunderstanding what you mean though.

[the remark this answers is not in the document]
joanibal: I'd feel a lot better about the PR if the step lengths were always real

joanibal (changes requested):
See comment about setting ANK_physLSTOl and other variables as always real times.

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
anilyil: Can you show me how you did it? Specifically, I want to have a pyADflow option that is a float, to be passed to an `alwaysRealType` variable in Fortran. I think when I changed the variable type to `alwaysRealType` both in the actual code and the pyadflow.pyf file, it _runs_ but it was passing random noise between the variables. The target variables in fortran ended up getting zeros most of the time if I remember correctly. Let me know if you know the solution to this issue and I can fix it. We can also make it a separate PR because there are actually a few variables that are supposed to be always real but we just use the real type because we set them from python.

tests/reg_tests/test_solver_combos.py
anilyil: I see, that makes sense. I will look into this. Worst case, we can have this case converge with 1 proc and that does converge to machine zero. I currently do not check derivatives with these tests but I am guessing you want to extend these tests to check the derivatives? 

As I previously said, I think checking total derivatives on a case where you know the numbers are reasonable makes more sense to me. So thats why I am fine with the adjoint/complex-step comparisons with the wing cases, and then this block test only looks at complex convergence. It is fine if you want to extend the test to complex step derivatives though.

anilyil: Well, thinking about the 2x2x2 cube again, does it even make sense to run that on 2 procs? We would end up getting two slices of the cube on each proc, but one of the directions would only have one layers. I am not sure if this would cause issues with the underlying double-halo communication and BC stuff. Should we just change this test to a 4x4x4 cube?

anilyil: So I just refined the 2x2x2 cube to get a 4x4x4 cube. All tests converge to machine zero w/o issues this way. So @sseraj's suggestion of doing a larger mesh is also satisfied this way. I will update the input_files archive and remove the custom l2 convergence for the Euler tests. 

I will probably rename the cube so that it is obvious that it is a 4^3 cube. I am guessing you will use the 2^3 cube for your own development @joanibal and I dont want to cause any confusion there. 

Does this make sense? Is this the last issue in this PR regarding the tests?

anilyil:
I updated the input files. The cube mesh is now a 4x4x4 block and the Euler tests also converge 12 orders of magnitude with 2 procs like the rest of the cases. Are there any other changes to the tests you want @joanibal ? I think now the last issue is to figure out the passing of alwaysRealType stuff from python to fortran?

src/NKSolver/NKSolvers.F90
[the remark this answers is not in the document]
anilyil: Okay so @joanibal implemented this `alwaysRealType` passing with f2py in his development branch. I created an issue detailing the fix here: https://github.com/mdolab/adflow/issues/133

@joanibal, I think we should do a separate PR for this since that possibly touches more than just these solvers. I think the solvers do work as is, so I think this PR should be fine. I agree with you though, it would be better if we could use `alwaysRealType` for these variables.

anilyil:
I think I addressed all of the comments with the tests and the code. The only thing I did not address yet is passing `alwaysRealType` options to Fortran with f2py. I created an issue for this https://github.com/mdolab/adflow/issues/133 and I think this should be a separate PR because it touches more places in the code. @joanibal and @sseraj are you happy with the current state? Do you want any other changes?

sseraj (approved):
Thanks for creating the issue. This is good to go for me.

joanibal (approved):

PR #136, opened 2021-03-28
Switch to using options list

nwu63:
## Purpose
This PR switches the options rendering from a table to a "list" of `data` directives. As part of this effort, you can now reference/link to specific options via ``:py:data:`<name>` ``.

Note that because I switched the YAML file from CRLF to LF line endings, the diff is huge (in fact the whole file). But because that's a separate commit, reviewers can simply look at the other commits to see what I changed there.

Also, there are probably other places where we talk about specific options but don't link to it, I will leave it to the maintainers to update that (feel free to push directly to this PR).

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)


sseraj:
Thanks! This is very clean. I will add some more links to options in the rest of the docs.

sseraj (approved):

anilyil (approved):
really good stuff. btw I could not get the links working on my local build of the docs. Is there sth I should be doing different or do we not expect them to work locally?

nwu63:
> really good stuff. btw I could not get the links working on my local build of the docs. Is there sth I should be doing different or do we not expect them to work locally?

No they should work, try updating the sphinx package with `pip install -U sphinx_mdolab_theme`, then rebuilding the docs.

anilyil:
yeah that did it. maybe I needed to update stuff.

PR #138, opened 2021-04-05
Fixed MExt

nwu63:
## Purpose
Previously `MExt` used to raise a lot of resource warnings regarding unclosed files. This PR fixes it by switching from `imp` to `importlib` since the former is deprecated. I also made some other minor adjustments to the code to improve readability/code quality.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Code that runs ADflow will not receive `ResourceWarning` anymore.

## Checklist

### DIFF
--- a/adflow/MExt.py
+++ b/adflow/MExt.py
@@ -1,5 +1,5 @@
 import tempfile
-import importlib
+from importlib.util import find_spec
 from pathlib import Path
 import os
 import shutil
@@ -15,7 +15,7 @@ def _tmp_pkg(tempDir):
     while True:
         path = tempfile.mkdtemp(dir=tempDir)
         name = os.path.basename(path)
-        spec = importlib.util.find_spec(name)
+        spec = find_spec(name)
         # None means the name was not found
         if spec is None:
             break
@@ -36,7 +36,7 @@ class MExt(object):
         self.name = libName
         self.debug = debug
         # first find the "real" module on the "real" syspath
-        spec = importlib.util.find_spec(packageName)
+        spec = find_spec(packageName)
         srcpath = os.path.join(spec.submodule_search_locations[0], f"{libName}.so")
         # now create a temp directory for the bogus package
         self._pkgname, self._pkgdir = _tmp_pkg(tmpdir)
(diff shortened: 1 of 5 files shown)


adflow/MExt.py
```diff
@@ -1,27 +1,28 @@
 import tempfile
-import imp
+import importlib
```
sseraj, line 2: When I try to test this locally, I run into this error: `AttributeError: module 'importlib' has no attribute 'util'`
Adding `import importlib.util` fixes the error. Did you run into this?

nwu63: No I didn't see this, very strange. What version of python are you running? Can you also check if `from importlib.util import find_spec` works? That may be more pythonic than your suggestion

sseraj: Yes, that also works. I'm running 3.8.5. The error does not show up when using `testflo` but does when running with `python`.

nwu63: Weird, I was testing some aerostructural scripts and didn't have any problems. I'm also on 3.8.5... anyways I will update the script. A quick google search shows that apparently importing the parent module does not automatically import sub-modules, so I guess in general it's always preferable to explicitly import the submodules that you want to use.

sseraj (approved):
Thanks. I tested this locally and the warning does not appear anymore.

nwu63:
Yeah [this SO answer](https://stackoverflow.com/questions/8696322/why-does-this-attributeerror-in-python-occur) explains what was happening. I didn't know this before but it's odd that it's non deterministic.

anilyil (approved):
Thanks for this work @nwu63. One question; would we also do a similar thing for other repos that use this approach?

nwu63:
> Thanks for this work @nwu63. One question; would we also do a similar thing for other repos that use this approach?

Yeah AFAIK it's only used in ADflow, IDwarp and pyHyp. I have a corresponding PR in IDWarp but I didn't think we needed to change pyHyp -- in the longer term I would like this file in baseclasses so we don't end up with code duplication.

PR #139, opened 2021-04-07
Various test enhancements

### LINKED ISSUE #110: A few enhancements for existing tests
# Description of feature
A few changes that should be adopted:
- ensure that `checkSolutionFailure` and `checkAdjointSolutionFailure` are run. Also assert true for training
- store metadata using the new API in baseclasses. At a minimum we should store all ADflow options, and maybe idwarp if it's used. Unless the values need to be retrained, we should only update the JSON files by appending a new key at the end.

marcomangano:
## Purpose
Closes #110 . I added two functions that wrap a check on the primal and adjoint solutions respectively - they work both for tests and training.
Now the training functions will also append the current ADflow options as metadata, and the `ref` files have been updated accordingly. I manually edited the file paths (`outputDirectory`, `gridFile`, ...) to remove my personal information.

The additions are pretty simple and straightforward, let me know if you have more elegant solutions in mind.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
I have tested that the test fail if `assertTrue(funcs['fail'])` is used.

## Checklist


nwu63 (dismissed):
This looks good, but it does beg the question of why we don't automatically call these `check` functions when `evalFunctions` or `evalFunctionsSens` are called... especially since they don't throw an error and just add the `fail` key to the dictionary. That would've reduced a lot of code duplication and it probably doesn't cost much. Thoughts? @anily @sseraj  @joanibal

sseraj:
> This looks good, but it does beg the question of why we don't automatically call these `check` functions when `evalFunctions` or `evalFunctionsSens` are called

I agree. This seems like the obvious thing to do.

anilyil:
Great additions @marcomangano. One question:

> I manually edited the file paths (`outputDirectory`, `gridFile`, ...) to remove my personal information.

So in the future, when people train the tests, do they also need to adjust these options? Also, it sounded like you manually modified the .json files, did I get that correctly? 

I think the proper way to do this is to avoid *any* manual edits to the json files. Your modifications do look super simple, but I would rather enforce this as a hard requirement for test reference files. 

If you can, it would be great to automate those two options when you are training the tests. I would almost say just not add them to the json because the input files etc are already clearly defined for all tests. That to me is the simplest solution. If you want to, you can also remove the directories automatically before the adflow path.

anilyil:
As a response to:

> This looks good, but it does beg the question of why we don't automatically call these check functions when evalFunctions or evalFunctionsSens are called... especially since they don't throw an error and just add the fail key to the dictionary. 

I would say there is still value in partially converged adjoint and flow solutions. If you prefer, you can add an option like they have in openmdao: https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/solver.py#L173 So ultimately this should be left to the user and should be easily customizable (if someone wants to keep moving on with a partially converged state, let them be), but adding the default checks and errors may help with new users getting used to checking for convergence etc.

anilyil (changes requested):
I am tagging this as requesting changes to prevent anyone else from merging this before the json stuff is addressed. To me, it should be a hard rule not to edit json files manually.

marcomangano:
> So in the future, when people train the tests, do they also need to adjust these options? Also, it sounded like you manually modified the .json files, did I get that correctly?
> 
> I think the proper way to do this is to avoid _any_ manual edits to the json files. Your modifications do look super simple, but I would rather enforce this as a hard requirement for test reference files.
> 
> If you can, it would be great to automate those two options when you are training the tests. I would almost say just not add them to the json because the input files etc are already clearly defined for all tests. That to me is the simplest solution. If you want to, you can also remove the directories automatically before the adflow path.

@anilyil Yeah I was thinking about removing the two options altogether, I don't think they are relevant for the tests. I trained the tests locally - maybe if we do so on a docker image we can keep that file location as a reference?

Anyway, I did not do manual additions, only edited the file paths. I trained the tests and then committed only the `metadata` entry, ignoring all the other modifications. Not ideal, but less invasive than an actual copy-paste. I agree that  we should not touch the files if possible... but this means that basically every PR on ADflow needs to be done through a docker image?

anilyil:
> but this means that basically every PR on ADflow needs to be done through a docker image?

No, just remove the stuff automatically (in the training script) so that the resulting json file can be used as is w/o *any* manual modification. I see you only changed the path etc, but still. The json should be fully automated. To me modifying a few innocent options by hand will just cause confusion to people who need to add new tests or train existing ones again. 

Also, you should commit full files after training if you re-train stuff in general, though for this case, it may make sense to just keep the metadata changes.

tests/reg_tests/reg_test_classes.py
```diff
@@ -31,5 +41,6 @@ def train(self):
         for test in tests:
             test_func = getattr(self, test)
             test_func()
+            self.handler.add_metadata(self.CFDSolver.options)
```
sseraj, line 44: I agree with Anil. There should be no manual changes. I think we should just pop the relevant options before adding them. (Also, use `getOptions` instead of directly accessing the options dictionary.)

nwu63:
Yeah we can look into either popping the entries we don't want, or converting them to relative paths or something. I've always used a Docker image in the past so this wasn't an issue. That being said, I don't agree with this
> Also, you should commit full files after training if you re-train stuff in general, though for this case, it may make sense to just keep the metadata changes.

I think we should only ever change the reference values if we expect the changes to affect the outputs, otherwise we should make sure to not touch them to avoid introducing unwanted drift over time.

> I would say there is still value in partially converged adjoint and flow solutions. If you prefer, you can add an option like they have in openmdao: https://github.com/OpenMDAO/OpenMDAO/blob/master/openmdao/solvers/solver.py#L173 So ultimately this should be left to the user and should be easily customizable (if someone wants to keep moving on with a partially converged state, let them be), but adding the default checks and errors may help with new users getting used to checking for convergence etc.

Right I see, that's a good point I forgot that `funcs['fail'] == True` will automatically stop a lot of the optimizations. That being said, I feel in most cases that's the behaviour we want, so we may want either an option, or for the user to just `pop` the `fail` key in their `objcon` or something if they want to override this.

marcomangano:
@anilyil @sseraj @nwu63 I committed the changes you suggested.
Can any of you try and retrain everything on docker? I have issues with the actuator region test training. Also, it's good to double check this before we merge

anilyil:
Will run the tests/training tomorrow at some point unless someone else does it before then.

tests/reg_tests/reg_test_classes.py
```diff
@@ -32,4 +43,14 @@ def train(self):
             test_func = getattr(self, test)
             test_func()
 
+            # Remove some path-related options before storing them as metadata
+            opts = copy.deepcopy(self.CFDSolver.getOptions())
```
sseraj, line 47: Two questions:
1) `getOptions()` already returns a copy of the dictionary. Did you need `deepcopy` to get it to work?
2) Does `KeyError` ever get raised?

nwu63: To add to this, it's preferable to just do `opts.pop(key, None)` as a one-liner, which will not raise `KeyError` if `key` is not found.

nwu63: Also, I would prefer to still keep in the paths, but convert them to relative paths (to the current working directory) so the personal/sensitive info is removed. I will give this a shot and push up soon.

nwu63:
Okay I pushed up some changes so now the paths are preserved but converted to relative paths first. I still had trouble with some actuator tests though, so could you take a look tomorrow @anilyil?

tests/reg_tests/reg_test_classes.py
marcomangano: Sure, I have no strong opinions about this, probably retaining the relative path makes more sense - additional and potential useful info. About your previous point, I wanted to delete the key altogether, doesn't that command keep the key with value `None`?

@sseraj: 1) I thought the same but I had some "missing key" errors while training.. apparently I was not popping out the copy? Pretty sure I was already using `getOption` but we can double check; 2) I guess it's extra safety... I realize just now that we are printing all the options, when I first manually deleted my personal folder path only some had `restartFile not None` so I got tricked. Anyway, if we opt to keep the relative path, this is not going to be an issue anymore.

marcomangano:
> Okay I pushed up some changes so now the paths are preserved but converted to relative paths first. I still had trouble with some actuator tests though, so could you take a look tomorrow @anilyil?

Ok I missed the last commit, I like this final code iteration! @nwu63  You haven't committed the updated `ref` files because of the actuator zone trouble?

tests/reg_tests/reg_test_classes.py
nwu63: Nope, `pop` actually is designed to return the value, and `None` will just specify that as the return value if `key` is not found, instead of raising the error. I also removed the extra copy just now -- I agree with @sseraj I don't think it's necessary.

nwu63:
> > Okay I pushed up some changes so now the paths are preserved but converted to relative paths first. I still had trouble with some actuator tests though, so could you take a look tomorrow @anilyil?
> 
> Ok I missed the last commit, I like this final code iteration! @nwu63 You haven't committed the updated `ref` files because of the actuator zone trouble?

Yeah I wanted @anilyil to take a look, one of the actuator zone tests require the complex ADflow to be built? Another failed for me locally, so I wanted to make sure it's not just me that was having that issue.

anilyil:
When I was writing the tests, I thought setup was called for each test method I defined under a class. Seems like when we are training, setup is called once, and the CFDSolver object is re-used for each test. As a result, I had to re-converge the CFD for a few of the tests. This fixed the inaccurate values in the actuator tests; however, now the json file is only updated after the first training runs, and then the second training method completely deletes all entries in the json file. I was using the same json file for multiple tests. Should I use different files for this or is it still possible to use the same json file?

nwu63:
> When I was writing the tests, I thought setup was called for each test method I defined under a class. Seems like when we are training, setup is called once, and the CFDSolver object is re-used for each test. As a result, I had to re-converge the CFD for a few of the tests. This fixed the inaccurate values in the actuator tests; however, now the json file is only updated after the first training runs, and then the second training method completely deletes all entries in the json file. I was using the same json file for multiple tests. Should I use different files for this or is it still possible to use the same json file?

Depends on how you have it set up. If they use the same test class, and you can get them to share the same CFDSolver + handler then it's possible. We can talk more offline if you want.

anilyil:
Yeah figured out my issue; it was due to the complex tests below using the same ref file. As I was fixing that, I realized another issue that is present with other tests. Basically, the `RegTest` class has a train method, and if complex tests inherit from `RegTest`, they also inherit the train method. As a result, if train is called on the whole file, the complex test's train is called after the real tests' train, which empties out the json file contents because complex tests dont add stuff to json files.

There is an ad-hoc implementation in adjoint tests to avoid this issue: https://github.com/mdolab/adflow/blob/master/tests/reg_tests/test_adjoint.py#L236 here the class that implements the complex tests inherit from `unittest.TestCase` instead of our `RegTest` to avoid having the `train` method. However, after this PR, we need the methods `assert_solution_failure` and `assert_adjoint_failure `. As a result, this PR adds these methods separately to each test script, which is code duplication: https://github.com/mdolab/adflow/pull/139/files#diff-0651f234b49859ad5551ca4ea22e37609c0dead29ca998788f51e9c26b064c9aR283-R292 

How I propose to solve this: we can create a CmplxRegTest class, that has the stuff we need but overloads the train method to just pass so we keep the json contents. That way, we can reduce the code duplication. 

I will push my changes in a second. Will post a follow up comment. I will become a contributor here after this so others should review my changes too :)

tests/reg_tests/test_actuator.py
```diff
@@ -252,9 +250,7 @@ def test_actuator_thrust_and_heat(self):
         az_heat = 1e5
         self.ap.setDesignVars({"thrust": az_force, "heat": az_heat})
 
-        # We dont need to rerun, restart file has the correct state. just run a residual
-        # self.CFDSolver(self.ap)
-        self.CFDSolver.getResidual(self.ap)
+        self.CFDSolver(self.ap)
```
marcomangano, line 253: Now that you switched to `__call__()` we can use `assert_solution_failure()` here and in the following sections as well!

marcomangano (commented):
I see now, makes sense! Please add the missing checks so everything is consistent.

anilyil:
I pushed my changes. See the newly added `CmplxRegTest` class, which solves the training issue with complex test classes. 

I have also fixed the actuator tests so they can be trained automatically, run fine, and training also does not require the complex adflow build as expected. 

I will re-run all tests and trainings locally to make sure everything is automated and works. I will post an update here once I do that. The PR is good to go for me besides that but others should review again.

anilyil:
I re-trained the tests, and only kept the changes in the metadata in each file. I did not commit the numerical changes, but besides those, the training works as I expect it to; everything is automatic. @marcomangano's changes so far is okay for me so I will approve this, but at least one person should review my changes.

anilyil (dismissed):

nwu63:
I took a look at your changes @anilyil and I like the new complex test class. I just made one minor change and replaced the overloaded class with a class attribute since we already had some skip training code there and this makes it a bit more explicit IMO. I also changed the paths to be relative to the root ADflow path so it doesn't depend on where you are running testflo from. Finally I updated the README for testing which was pretty out of date. This looks good to me so if you are happy with the changes I think we can merge.

nwu63 (approved):

anilyil (approved):
Thanks for the work everyone

PR #140, opened 2021-04-29
Updated liftIndex docs

sseraj:
## Purpose
As far as I can tell, there is no automatic check for the lift direction and no way to leave `liftIndex` unspecified. I removed these lines from the docs.

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/doc/options.yaml
+++ b/doc/options.yaml
@@ -294,8 +294,6 @@ restrictionRelaxation:
 liftIndex:
   desc: >
     Specify the coordinate index that will be considered the lift direction.
-    If not supplied, this parameter will be determined automatically if there are symmetry planes present in the grid.
-    Otherwise, it must be supplied.
   2: y-axis
   3: z-axis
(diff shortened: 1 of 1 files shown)


anilyil:
if you run it w/o specifying the lift index, does the code tell you that you have to specify it? if so, I am fine with this PR.

sseraj:
Not specifying `liftIndex` in the options will just mean that it defaults to 2. Looking at the git history, this has been the default behavior since at least [this commit](https://github.com/mdolab/adflow/commit/e35caeb932c20fa9b8d88077d32f143196956b6d) in 2014. I don't know if this is the best approach, but it should be fine as long as the documentation is not misleading.

anilyil (approved):

joanibal (approved):

PR #152, opened 2021-07-27
Number the failed mesh output

nwu63:
## Purpose
This PR adds numbering to `failed_mesh.cgns` so that multiple failed meshes can be written out without overwriting the previous one.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I verified this on Great Lakes and it works as expected.

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -173,6 +173,7 @@ class ADFLOW(AeroSolver):
         self.updateTime = 0.0
         self.nSlice = 0
         self.nLiftDist = 0
+        self.nMeshFail = 0
 
         # Set default values
         self.adflow.inputio.autoparameterupdate = False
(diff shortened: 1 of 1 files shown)


marcomangano (approved):
Very good, should have done this earlier!

sseraj (approved):

adflow/pyADflow.py
```diff
@@ -1040,14 +1041,15 @@ def __call__(self, aeroProblem, **kwargs):
                 )
 
         if self.adflow.killsignals.fatalfail:
-            print("Fatal failure during mesh warp! Bad mesh is " "written in output directory as failed_mesh.cgns")
-            fileName = os.path.join(self.getOption("outputDirectory"), "failed_mesh.cgns")
-            self.writeMeshFile(fileName)
+            fileName = f"failed_mesh_{self.nMeshFail:02}.cgns"
```
joanibal, line 1044: I think it would be useful to number the mesh failures using the normal call counter. That way you can easily where in the iteration history the failed mesh goes.

nwu63: Oh yeah this is actually a good point. Would you mind opening a subsequent PR to patch this?

PR #153, opened 2021-07-28
Sa rough

DavidAnderegg:
## Purpose
This is a hacky implementation of the SA-rough model as described [here](https://turbmodels.larc.nasa.gov/spalart.html#sarough). The goal of this PR is to collect the changes in a central place so other ADflow users might expand on it. I talked to @anilyil about this plan.

## Implementation
I introduced a new variable in the python wrapper. It is called `kssa`and controls the `equivalent sand grain roughness` as described above. This roughness gets applied to all wall surfaces. 
It also only works with deactivated blockettes `useblockettes: False`.

## Issues
I had some problems running this code on our cluster where it would randomly crash. I thought this might be due to a division by 0 when calculating the SA-factor close to a wall. So i implemented some checks for this in the `safact` function (`src/turbulence/turbBCRoutines.F90`). But this seems not to have solved it back then.

I only experienced these crashes on our cluster, but never locally. They also seem to have stopped after a while. I have no clue what is going on. This might be a cluster specific problem.

## Validation
I did no proper validation, but I compared something with the literature and made sure the optimization would still work. You find all my scripts [here](https://github.com/mdolab/adflow/files/6893115/sa_rough_tests.zip).

### model validation
I simulated the NACA 63348 airfoil with various ks-values and plottet it against the measurements from this [report](https://energy.sandia.gov/wp-content/uploads/2017/10/LEE_Wilcox_SAND2017-11288.pdf) (page 55). This is somewhat of an apples to oranges comparison as i could only apply the roughness to the whole surface and in the mentioned report only the LE was contaminated.
n63348_sa_rough

### gradient validation
In order to validate the gradients, i create a really simple FFD-box around a NACA0012 airfoil with only 8 FFD-Points (one on each point spanning the whole airfoil). Then i calculated the gradients using the adjoint version on the clean airfoil (ks=0) and contaminated airfoil (ks=5e-4) and compared it with the gradients of the contaminated airfoil using finite differences:
adjoint

### optimization validation
In order to make sure the optimization would work, i optimized a NACA0012 airfoil at alpha=0 degrees for drag (this was probably not the smartest idea...) under different ks-values. This is the result:
opt_overview
opt_shape

### conclusion
As you can see, it does somewhat work but I wouldn't bet my house on the correctness. I used it only as an indication for roughness sensitivity. For this, it was good enough.

## Steps needed for a proper implementation
For a proper implementation those steps are needed:

- Find a way to apply different ks-values to different surfaces (i would go for the surfacefamily approach)
- Implement blockettes
- Validate it 

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist


DavidAnderegg:
I do realize there is a lot of commented code. I will clean this up.

anilyil (approved):

anilyil:
These changes look good to me for now since we are merging this to a separate branch anyways. Thanks for sharing this work @DavidAnderegg. @marcomangano I suggest you keep track of this branch after this point since you are the main person who might benefit from this in the future. 

@nwu63, the docs fail because this is a different branch probably? Can we override that and merge this?

nwu63:
> @nwu63, the docs fail because this is a different branch probably? Can we override that and merge this?

Yeah exactly, but it's not a required check here since this is not the protected `master` branch. We can just merge with 2 approvals, or I can just merge it if you want.

anilyil:
Either way is fine. I can also merge if you approve. I think the changes are minimal and a good starting point if we want to officially support this in the future. I think I updated the sa-rough branch with mdolab/master recently, so that should be good as well.

nwu63:
I'll let @marcomangano or @sseraj approve/merge this, since they are actual maintainers and this stuff is somewhat relevant for them.

Issue #157, opened 2021-08-16
some cgns surf variables incorrectly written to file when cgns zone subfaces are split and using storeRind 


joanibal:
## Description
 
When including the "rind" cells, the output range for each flowDom block is widened to include the rind if it includes the edge cells. 
If a flowDom spans the entire i range of a cgnsblock (i.e. the cgns zone has not been split in the i direction) with a size of 10,

for example 
cellRange(1,:) = 2 ,9
will become 
cellRange(1,:) = 1,10
when the rind cells are included 

However not all  the variables used for the output are defined in the rind cells.  In particular the values for the viscSubfaces do not include the rind cells 
https://github.com/mdolab/adflow/blob/e091afce4766eaf9a6b7061fda33369ed09f6629/src/preprocessing/preprocessingAPI.F90#L2547-L2551

So additional logic in the output subroutine is used to output the values of the next interior cell instead of the rind cell. 
It does this by comparing the current index to the min and max range of indices given for that flowDom block subface. 
When the index is equal to a min or a max it increases or decreases it by one respectively. 
https://github.com/mdolab/adflow/blob/e091afce4766eaf9a6b7061fda33369ed09f6629/src/output/outputMod.F90#L1885-L1901

So for a cellRange(1,:) = 1,10 
i = 2,2,3,4,5,6,7,8,9,9 
will be output 

However when the cgns zone is split,  the min and max of a flowDom no longer necessarily correspond to rind cells.

for example, if the i range is split into three.
 cellRange(1,:) = 2,4 
 cellRange(1,:) = 5,6 
 cellRange(1,:) = 7,9
After including the rind cells become 
 cellRange(1,:) = 1,4 
 cellRange(1,:) = 5,6 
 cellRange(1,:) = 7,10
 and will output
i = 2,2,3,3  6,5  8, 8, 9,9

data for i = 4, 7 are missing and some data has been repeated unnecessarily and some flipped. 

### Steps to reproduce issue
For a problem with a surface output dependent on viscSubFace

1. add the option "partitionLikeNProc": 50(may need to be higher if you have a finer mesh),
2. try to write a surface solution file
3. may get bad output or seg fault

PR #158, opened 2021-08-16
fixed output ranges for split visc subfaces

joanibal:
## Purpose
The goal of this PR is to address is issue #157 
See the issue for details on the issue. 
TLDR: when the subfaces of viscous zones were split, the output data was slightly wrong 

## Type of change

## Testing
For a problem with a surface output dependent on viscSubFace
1.   add the option "partitionLikeNProc": 50(may need to be higher if you have a finer mesh),
2.   try to write a surface solution file
3.   check for bad output or seg fault

## Checklist

- [NA] I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted

### DIFF
--- a/src/output/outputMod.F90
+++ b/src/output/outputMod.F90
@@ -1372,8 +1372,9 @@ contains
   end subroutine storeSolInBuffer
 
   subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
-       faceID, cellRange, solName, &
-       viscousSubface, useRindLayer)
+                                  faceID, cellRange, solName, &
+                                  viscousSubface, useRindLayer, &
+                                  iBeg, iEnd, jBeg, jEnd)
     !
     !       storeSurfsolInBuffer stores the variable indicated by
     !       solName of the given block ID in the buffer. As the solution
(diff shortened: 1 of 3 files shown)


joanibal:
Before: 
Screenshot from 2021-08-16 14-31-35

After:
Screenshot from 2021-08-16 14-30-38

anilyil:
Good stuff. I will go through the code and try to understand it a bit. Is there a way we can test this?

src/output/outputMod.F90
```diff
@@ -1881,25 +1884,42 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
...
-                ii = min(i + offVis, rangeFace(1,2))
-             else if(i == rangeFace(1,2)) then
-                ii = max(i - offVis, rangeFace(1,1))
-             else
-                ii = i
-             endif
+         !  if satements are used to copy the value of of the interior
+         ! cell since the value isn't defined in the rind ell 
 
+         if (present(jBeg) .and. present(jEnd) .and. (useRindLayer)) then
```
anilyil, line 1892: Why did you define i/j beg/end as optional input variables here? Is this routine used anywhere else in adflow?

src/output/outputMod.F90
```diff
@@ -1881,25 +1884,42 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
...
+         else
+            jj = j
+
+         end if
+
+          do i=rangeFace(1,1), rangeFace(1,2)
+             if (present(iBeg) .and. present( iEnd) .and. (useRindLayer)) then 
+               ior = i + iBegor - 1
+               if (ior == iBeg) then 
+                  ! print *, 'ibeg'
```
anilyil, line 1910: Can you remove the print statements?

src/output/writeCGNSSurface.F90
```diff
@@ -1320,7 +1320,8 @@ subroutine writeSurfaceSol
                call storeSurfsolInBuffer(ind, buffer, jj, ii,       &
                     faceID, cellRange(1,1,kk), &
                     solNames(mm),              &
-                    viscousSubface, storeRindLayer)
+                    viscousSubface, storeRindLayer,&
+                    iBeg, iEnd+1, jBeg, jEnd+1)
```
anilyil, line 1324: Is there a reasoning (other than convenience) for adding the 1s here to i/j end here instead of the low level checks in the subroutine itself? The additions here can cause confusion; the i/j beg/end variables have the same names here and the subroutine you are calling here, but the values will be different. I would rather put the +1s right when the variables are actually used, and maintain the same values for these variables which have the same names. Just a convenience thing.

anilyil (changes requested):
Thanks for fixing this, and also the detailed explanation in #157. I have left a few comments.

PR #161, opened 2021-08-30
Flake8 fix

nwu63:
## Purpose
This PR fixes the flake8 violations as a result of the updated configurations. I decided it was probably not worth fixing some of the files, so I added them to the flake8 ignore list.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist


tests/reg_tests/test_adjoint.py
```diff
@@ -54,7 +54,7 @@ def twist(val, geo):
     def span(val, geo):
         # Span
         C = geo.extractCoef("wing")
-        for i in range(len(C) - 1):
+        for _i in range(len(C) - 1):
```
nwu63, line 57: I have no idea whether the original code was correct since I've not used the span function before. Why do we not use the loop variable here?

adflow/pyADflow.py
```diff
@@ -5254,25 +5254,23 @@ def _getSpecialOptionLists(self):
         # pythonOptions do not get set in the Fortran code.
         # They are used strictly in Python.
 
-        pythonOptions = set(
```
marcomangano, line 5257: Why did you change this from a set to a "flat" dictionary?

tests/reg_tests/test_adjoint.py
marcomangano: This looks like a bug to me, we should move all the `refAxis` points, the loop is clearly useless otherwise

marcomangano (commented):
How much effort would it be to fix the other files you added to `extend-exclude`? I agree they look low priority, but we might want to fix them once for all at some point?

adflow/pyADflow.py
nwu63: They are the equivalent syntax: `{'a', 'b'} == set(('a', 'b'))`.

tests/reg_tests/test_adjoint.py
nwu63: Yeah that was my thought also, but I don't really know how to fix it. In any case this highlights how checking unused loop variables can prevent bugs.

marcomangano: @joanibal it looks like you were involved in #47 when we updated the testing section, do you recall how is this DV supposed to work?

nwu63:
> How much effort would it be to fix the other files you added to `extend-exclude`? I agree they look low priority, but we might want to fix them once for all at some point?

They are not too bad, but other than `fort_depend` and `autoEdit` the others are probably going to be deprecated soon (if not already) so I didn't feel like fixing them.

marcomangano:
> They are not too bad, but other than `fort_depend` and `autoEdit` the others are probably going to be deprecated soon (if not already) so I didn't feel like fixing them.

Ok I see, let's just don't forget about them so when they'll be removed we can also clean up the few remaining files

marcomangano (approved):

marcomangano:
@joanibal @ArshSaja @sseraj could any of you check this and merge if there are no issues?
Also please take a look at the potential "bug" in `test_adjoint` I commmented above

tests/reg_tests/test_adjoint.py
sseraj: I agree that this is a bug. The span variable should be something like:
```
        def span(val, geo):

            C = geo.extractCoef("wing")
            s = geo.extractS("wing")

            for i in range(len(C)):
                C[i, 2] = C[i, 2] + s[i] * val[0]

            geo.restoreCoef(C, "wing")
```

However, I don't think this is affecting the test coverage and fixing it would require retraining the test. I can create an issue and then merge this in as is.

marcomangano: Agreed @sseraj

sseraj: The issue is at #163

sseraj (approved):

tests/reg_tests/test_adjoint.py
joanibal: I have no explanation for this. It was probably a temporary change for debugging that accidentally became permanent.

PR #168, opened 2021-10-13
Added an option to specify acceptable adjoint convergence

sseraj:
## Purpose

I added an option called `adjointMaxL2DeviationFactor` that allows the user to specify how much worse than the target convergence is acceptable for the adjoint solution. This works the same way as `maxL2DeviationFactor` but for the adjoint instead of the flow solution.

This is a more precise alternative to setting `skipAfterFailedAdjoint` to False. See #45 for some related discussion.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/output/outputMod.F90
+++ b/src/output/outputMod.F90
@@ -1372,8 +1372,9 @@ contains
   end subroutine storeSolInBuffer
 
   subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
-       faceID, cellRange, solName, &
-       viscousSubface, useRindLayer)
+                                  faceID, cellRange, solName, &
+                                  viscousSubface, useRindLayer, &
+                                  iBeg, iEnd, jBeg, jEnd)
     !
     !       storeSurfsolInBuffer stores the variable indicated by
     !       solName of the given block ID in the buffer. As the solution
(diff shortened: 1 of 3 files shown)


sseraj:
> did the test fail randomly again?

Yes, it did. I'll rerun the tests for this PR and also try to prioritize fixing the memory issue. It seems like the failures are showing up more frequently now somehow, possibly due to changes on the Azure runners.

nwu63:
Why did codecov not work here? The logs say 
```
Error POSTing to https://codecov.io: 500 {"error": "Server Error (500)"}
There was an error running the uploader: Error uploading to https://codecov.io: Error: Internal Server Error
```

sseraj:
> Why did codecov not work here?

PRs from forks (no token) still run into this issue but there is an open PR on codecov to fix this (codecov/uploader#426).

nwu63:
Wow that was immediate. Do we wanna remove all the codecov tokens and stuff @sseraj

sseraj:
> Wow that was immediate. Do we wanna remove all the codecov tokens and stuff @sseraj

I don't think there's any downside to keeping the tokens, especially because we need the variable group anyway. But I can remove them if you prefer.

Issue #133, opened 2021-03-19
Convert some pyADflow options to alwaysRealType to get simpler code in complex mode.


anilyil:
# Description of feature

Currently we can only pass pyADflow options that are defined as `float` in Python to the `realType` in Fortran. This is because of the limitation of the current state of the .pyf preprocessor. Ideally, there are several python-level options that should be passed as `alwaysRealType` due to how they are used inside Fortran. For most of these, we have extra `real()` checks in Fortran and complex mode, but if we passed these to Fortran as `alwaysRealType`, we would be able to remove these checks. Furthermore, this leaves some room for error where the float variables may result in a non-zero complex part, though I never ran into this up to now.

# Potential solution

The pyf_preprocessor.py file at https://github.com/mdolab/adflow/blob/master/src/f2py/pyf_preprocessor.py is responsible from parsing the preprocessor. We need to add the case for `alwaysRealType` in the check here: https://github.com/mdolab/adflow/blob/master/src/f2py/pyf_preprocessor.py#L75 

@joanibal has an implementation of this in his dev branch: https://github.com/mdolab/adflow/compare/master...joanibal:dev#diff-2674555147d45ec975151530704950e671ebb3e8f70d0333a2bee0af06ad0750 

The modification of the relevant part of the code looks like this:

```
            if 'real(kind=realtype)' in orig_lines[i]:
                if mode == 'real':
                    orig_lines[i] = orig_lines[i].replace('kind=realtype','kind=8')
                else:
                    orig_lines[i] = orig_lines[i].replace(
                        'real(kind=realtype)','complex(kind=8)')

            if 'real(kind=alwaysrealtype)' in orig_lines[i]:
                orig_lines[i] = orig_lines[i].replace('kind=alwaysrealtype','kind=8')
```

After we make this change to the pyf preprocessor, we also should update all of the options used as always real but defined as `realType` in the Fortran level to use the `alwaysRealType` instead of the `realType`. These include options like convergence tolerances, CFL numbers, etc.

PR #174, opened 2021-11-15
Tools for Debugging Jacobian Vector Products and some small bug fixes

joanibal:
## Purpose
Adds some tools for debugging Jacobian vector products such as a finite difference and complex step mode for the forward mode.
It also added a wrapper for the state only jac vec prods used for the adjoint.  

The PR also has 3 minor bug fixes or changes not directly related to this:
1.  The vec get/restore stuff for `xSurfVec` in the master routine.
2. Changing the order of calls with `timestep_block` in the master routines.
3. The `alwaysRealType` check in the pyf preprocessor, which addresses part of #133.

The calls to `timestep_block` in the master, master_d, master_b were moved to a consistent location within each. 

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I'm added some test for the feature, but I'm not totally satisfied with them.
We need a better way to directly compare the output of the FWD and FWD with CS. 
You could do it with a reference file, but that would require retraining even when both change. 

## Checklist


src/adjoint/adjointDebug.F90
```diff
@@ -0,0 +1,736 @@
...
+        end do
+
+
+        if(equations == RANSEquations) then
+            call updateXSurf(level)
+        end if
+        bcDataValues = bcDataValues - bcDataValuesdot*h
+
+        
+        &
```
SichengHe, line 281: & symbol a typo?

src/adjoint/adjointDebug.F90
```diff
@@ -0,0 +1,736 @@
+! This module is used for debugging and testing only
+
+module adjointDebug
+
+contains
+
+#ifndef USE_COMPLEX
+
+    subroutine computeMatrixFreeProductFwdFD(xvdot, extradot, wdot, bcDataValuesdot,&
```
SichengHe, line 9: Overall, it feels that this shall be located in the same file with computeMatrixFreeProductFwd and computeMatrixFreeProductBwd similar with master to master_d and master_b because the function is very similar with computeMatrixFreeProductFwd.

src/adjoint/adjointDebug.F90
```diff
@@ -0,0 +1,736 @@
...
+
+        ! Input parameters
+        integer(kind=intType) :: nn, level, sps
+
+        ! Working parameters
+        integer(kind=intType) :: mm, i, iDom
+        integer(kind=intType) :: iBoco, iData, iDirichlet
+        write(*,*) 'ADSeeds for block', nn, ' at level ', level, ' at sps ', sps
+        write(*,*) 'd2wall ', minval(flowDomsd(nn, level, sps)%d2wall), &
+                        maxval(flowDomsd(nn, level, sps)%d2wall)
```
SichengHe, line 315: Besides the max and min, I think the L2 norm gives some idea of the overall match of the data.

src/adjoint/adjointDebug.F90
```diff
@@ -0,0 +1,736 @@
...
+            write(*,*) 'actuatorRegionsd(i)%Force ',actuatorRegionsd(i)%force
+            write(*,*) 'actuatorRegionsd(i)%Torque ',actuatorRegionsd(i)%torque
+        end do
+
+    end subroutine printADSeeds
+
+
+#else
+
+    subroutine computeMatrixFreeProductFwdCS(xvdot, extradot, wdot, bcDataValuesdot, &
```
SichengHe, line 523: Similar with FD, I feel it shall go to adjointAPI.f90.

src/adjoint/masterRoutines.F90
```diff
@@ -94,14 +98,24 @@ subroutine master(useSpatial, famLists, funcValues, forces, &
           call setPointers(nn, 1, sps)
 
           if (useSpatial) then
+
+             call VecGetArrayF90(xSurfVec(1, sps), xSurf, ierr)
```
SichengHe, line 102: This is a good catch. I had it fixed a branch. It is nice it is fixed.

tests/reg_tests/test_jacvecprodBWDFast.py
```diff
@@ -0,0 +1,100 @@
...
+            resBar=dwBar,
+            wDeriv=True,
+        )
+
+        wBarfast = self.CFDSolver.computeJacobianVectorProductBwdFast(resBar=dwBar)
+
+        np.testing.assert_allclose(wBar, wBarfast, atol=1e-16, err_msg="w wrt res")
+
+    def test_repeated_calls(self):
+
```
SichengHe, line 89: Why do we need this test?

SichengHe (commented):
Overall I feel this pr is very useful for debugging AD code!! 
Currently, when we debug the FD with FAD, we need to store the seeds and compare FD and FAD at the corresponding locations from master and master_d. This requires quite some recompiling. I am thinking is it possible to make this process automatic or at least compiling-free?

tests/reg_tests/test_jacvecprodBWDFast.py
joanibal: In the reverse mode the primal variables can be modified, which can cause the value of the derivatives to change after consecutive calls. 
This has been a problem for me in the past.

joanibal:
I'm no sure I have a good way to fix the recompiling problem. 
We talked about adding some  preprocessor statements to the master routines to print all the AD seeds at different points.
Something like that could be useful, but at the expense of greatly cluttering the code. 
Additionally, after you have narrowed down the problematic variable you usually just want to print that one rather than all of them. 
I'm open to ideas about how to fix this, but that could always come in a separate PR.

As for moving the debugging routines in the `adjointAPI` file. 
I like having them separated, but I'd like to hear what the other reviewers think too.

joanibal:
This PR is ready for a review by either @marcomangano, @anilyil , or @eirikurj

adflow/pyADflow.py
```diff
@@ -3738,6 +3738,8 @@ def computeJacobianVectorProductFwd(
         funcDeriv=False,
         fDeriv=False,
         groupName=None,
+        mode="AD",
```
anilyil, line 3741: Can you add the explanation of mode and h to the doctoring? It would help if you explain what the options are for mode, and how they all work. 

Also, I am conflicted by the default value for h; its only good for FD, its not used in AD and should be modified for CS. Is there a better way we can specify this? Maybe have none as default and each case can overwrite its own default unless the user specifies the value they want?

adflow/pyADflow.py
```diff
@@ -4135,6 +4178,26 @@ def computeJacobianVectorProductBwd(
         # Single return (most frequent) is 'clean', otherwise a tuple.
         return tuple(returns) if len(returns) > 1 else returns[0]
 
+    def computeJacobianVectorProductBwdFast(
+        self,
+        resBar=None,
+    ):
+        """used for testing the routines used for jac vec prods for the adjoint solve.
```
anilyil, line 4185: Can you add more explanation here? "adjoint solve" is the type of language that does not really explain much while confusing newcomers. These fast routines are bwd-differentiated residual routines only with respect to the state and residuals themselves. We can also mention that this is the operator used for the matrix-free solution of the adjoint system.

adflow/pyADflow.py
```diff
@@ -4135,6 +4178,26 @@ def computeJacobianVectorProductBwd(
...
+        ----------
+        resBar : numpy array
+            Seed for the residuals (dwb in adflow)
+
+        Returns
+        -------
+        wbar: array
+            state derivative seeds
+        """
+        wbar = self.adflow.adjointapi.computematrixfreeproductbwdfast(resBar, self.getAdjointStateSize())
```
anilyil, line 4197: I am guessing there is a prerequisite that the intermediate variables need to be up to date before calling this? Can you provide a small guide type documentation on how to use these routines for debugging?

src/adjoint/adjointAPI.F90
```diff
@@ -2,8 +2,10 @@ module adjointAPI
 
 contains
 #ifndef USE_COMPLEX
-  subroutine computeMatrixFreeProductFwd(xvdot, extradot, wdot, bcDataValuesdot, useSpatial, &
-       useState, famLists, bcDataNames, bcDataValues, bcDataFamLists, bcVarsEmpty, dwdot, funcsDot, fDot, &
+  subroutine computeMatrixFreeProductFwd(xvdot, extradot, wdot, bcDataValuesdot,&
+       useSpatial, useState, famLists,&
+       bcDataNames, bcDataValues, bcDataFamLists, bcVarsEmpty,&
+       dwdot, funcsDot, fDot, &
```
anilyil, line 8: is there a specific reason for the change here? Looks like just formatting?

src/adjoint/adjointAPI.F90
```diff
@@ -232,6 +235,11 @@ subroutine computeMatrixFreeProductBwdFast(dwbar, wbar, stateSize)
     ! of the fact that only wbar needs to be zeroed since all other
     ! required seeds are zeroed in the individual fast routines. This is
     ! slightly unsafe, but it necessary for speed.
+    
+    ! Allocate the memory for reverse
```
anilyil, line 239: is this only used for the debug functions? would it ever interfere with the regular use-cases?

src/adjoint/masterRoutines.F90
```diff
@@ -1,7 +1,9 @@
 module masterRoutines
 contains
-  subroutine master(useSpatial, famLists, funcValues, forces, &
-       bcDataNames, bcDataValues, bcDataFamLists)
+  subroutine master(useSpatial,&
+                  famLists, funcValues,&
+                  forces, &
+                  bcDataNames, bcDataValues, bcDataFamLists)
```
anilyil, line 6: again, this looks like a formatting change, correct? Is there a specific styling you were going for or is this just random? I would suggest not making modifications like this unless we fix entire modules' syntax at a time. Otherwise its just adding confusion and extra lines to the PR.

src/adjoint/masterRoutines.F90
```diff
@@ -94,14 +98,24 @@ subroutine master(useSpatial, famLists, funcValues, forces, &
           call setPointers(nn, 1, sps)
 
           if (useSpatial) then
+
+             call VecGetArrayF90(xSurfVec(1, sps), xSurf, ierr)
+             call EChk(ierr,__FILE__,__LINE__)
+
```
anilyil, line 104: Is this a bug fix? Is there something I need to check here?

src/adjoint/masterRoutines.F90
```diff
@@ -144,6 +158,7 @@ subroutine master(useSpatial, famLists, funcValues, forces, &
     do sps=1, nTimeIntervalsSpectral
        do nn=1, nDom
           call setPointers(nn, 1, sps)
+          call timeStep_block(.false.)
```
anilyil, line 161: Why did you move this call here? We would need to also adjust the master_d and master_b routines to reflect the correct order because they are hand-differentiated.

src/adjoint/masterRoutines.F90
```diff
@@ -38,6 +40,8 @@ subroutine master(useSpatial, famLists, funcValues, forces, &
     use inputOverset, only : oversetUpdateMode
     use oversetCommUtilities, only : updateOversetConnectivity
     use actuatorRegionData, only : nActuatorRegions
+    use wallDistanceData, only : xSurfVec, xSurf, wallScatter
```
anilyil, line 43: do we use `wallScatter` here?

src/f2py/pyf_preprocessor.py
```diff
@@ -69,12 +69,16 @@
...
                 orig_lines[i] = orig_lines[i].replace("kind=inttype", "kind=4")
+
             if "real(kind=realtype)" in orig_lines[i]:
                 if mode == "real":
                     orig_lines[i] = orig_lines[i].replace("kind=realtype", "kind=8")
                 else:
                     orig_lines[i] = orig_lines[i].replace("real(kind=realtype)", "complex(kind=8)")
-                # end if
+
+            if "real(kind=alwaysrealtype)" in orig_lines[i]:
```
anilyil, line 79: is this the fix required to have always real type options passed from python? does this require any other changes to the code?

tests/reg_tests/test_jacvecprodBWDFast.py
anilyil: why do the primal variables get modified in repeated bwd calls?

tests/reg_tests/test_jacvecprodBWDFast.py
```diff
@@ -0,0 +1,100 @@
+# built-ins
```
anilyil, line 1: Can you change the test file name so that the part after the test is camel case (or whatever format you prefer)? so `test_jacVecProdBWDFast.py` maybe?

tests/reg_tests/test_jacvecprodFWD.py
```diff
@@ -0,0 +1,214 @@
+# built-ins
```
anilyil, line 1: Can you also change the filename here in line with what I wrote down for the BWD version?

tests/reg_tests/test_jacvecprodBWDFast.py
```diff
@@ -0,0 +1,100 @@
...
+        "aero_prob": copy.deepcopy(ap_tutorial_wing),
+        "N_PROCS": 1,
+    },
+]
+
+
+@parameterized_class(test_params)
+class TestJacVecBWDFast(reg_test_classes.RegTest):
+    """
+    Tests that given a flow state the residuals, function, forces/tractions,
```
anilyil, line 42: is this what this test checks or does this explanation need to be updated?

tests/reg_tests/test_jacvecprodFWD.py
```diff
@@ -0,0 +1,214 @@
...
+            return
+
+        # perturb each input and check that the outputs match the FD to with in reason
+        wDot = self.CFDSolver.getStatePerturbation(321)
+
+        resDot_CS, funcsDot_CS, fDot_CS = self.CFDSolver.computeJacobianVectorProductFwd(
+            wDot=wDot, residualDeriv=True, funcDeriv=True, fDeriv=True, mode="CS", h=self.h
+        )
+
+        print(resDot_CS)
```
anilyil, line 210: Do we compare this again anything? I remember there was a discussion on using json files for complex tests as well. I think I defaulted to comparing against the values out of the real tests. Maybe we can do a similar thing here.

src/adjoint/adjointDebug.F90
anilyil: I am fine with having it here as long as all functions here are for debugging and never used for actual production runs. Plus this is probably easier to include these in a complex compile, I think the other adjoint stuff is not included in complex build.

src/adjoint/adjointDebug.F90
```diff
@@ -0,0 +1,735 @@
...
+
+        else
+            call master(useSpatial, &
+                        famLists, funcValues, &
+                        forces, &
+                        bcDataNames, bcDataValues, bcDataFamLists)
+        end if
+      
+
+        ! Copy basevale (f) into variables for the final vals (f(x+dx) - f)/dx
```
anilyil, line 123: typo? "basevale"

src/adjoint/adjointDebug.F90
anilyil: I agree with @SichengHe here, probably worth doing the L2. 

Also, do you anticipate using these routines in parallel runs? How would these print statements work there? In that case, would we distinguish each value from each proc, or would we report the global values? Printing different stuff from procs w/o paying attention to the order will break the "use it with a diff tool" usecase.

src/adjoint/adjointDebug.F90
anilyil: I say we keep it here because that module is not even included in the complex builds whereas we want this to be included.

anilyil (changes requested):
Thanks for putting in the work @joanibal. I have left my comments. I will do another review after we address the first set to make sure I did not miss anything.

src/adjoint/adjointAPI.F90
joanibal: Yes, `solveAdjointForRHS` allocates the memory first during typical use. 
Since it checks that the memory is not allocated first there is no chance of double allocation.

src/adjoint/masterRoutines.F90
joanibal: no

src/adjoint/masterRoutines.F90
joanibal: It is a bug fix. 
But, this part of the master routine is not run during normal operations. 
You can see that to be consistent with the FWD routine, this must be there.

tests/reg_tests/test_jacvecprodBWDFast.py
joanibal: In general, because if the value of a primal variable is modified in the forward pass, that modification must be undone in the reverse pass so the correct intermediate values are used. But repeated undoings of modifications can cause problems. 

I don't remember entirely what my specific problem was. 
Based on the comment I added to my fork. The issue was related to the halo cell states being changed on each update.

joanibal: At the end of the day, it doesn't hurt to check

src/adjoint/adjointDebug.F90
joanibal: L2 added.

I have never used this routine in parallel. 
You'd probably want to redirect the output of each proc to a separate file  (which can be done with a flag).

adflow/pyADflow.py
joanibal: I changed the default value of h to None

src/adjoint/masterRoutines.F90
joanibal: This has been undone

adflow/pyADflow.py
joanibal: The intermediate variables do need to be up to date. 
The tests provide an example of how to do this (just call get residuals before).

When this gets accepted I'll update [these docs](https://mdolab-private-docs.readthedocs-hosted.com/en/latest/tutorials/debugging.html#debugging-derivatives)

anilyil: Sounds good

tests/reg_tests/test_jacvecprodBWDFast.py
anilyil: ok, I buy that

tests/reg_tests/test_jacvecprodFWD.py
anilyil: > However, then every time we make a change to the primal code this test would fail.
Then you would re-train at that point.

> It would be preferable to call the AD routines in complex mode and directly compare the values with CS.
Can you create a real ADflow instance and test CS against AD here? 

> In practice, is it possible to complexity the ADed routines?
Yeah its doable but I don't get why that's useful here. doing CS over AD kinda gives you second derivatives. Plus its much more pain than what you are trying to do here.

src/adjoint/adjointDebug.F90
anilyil: Okay, then as long as you also include these use cases in the documentation update, its fine.

src/adjoint/masterRoutines.F90
anilyil: ok, the order is consistent with others. I am curious why this was not updated here before. 

I think I also remembered why we were struggling with this before; when we did the blockette stuff, we weren't putting back the radI/J/K variables from blockette memory, which was causing issues with other residual evaluations.

either way, the change in this PR looks fine.

anilyil:
I left a few comments and tried to resolve the outstanding ones. Let me know if you have any questions. I think the last remaining comment I have is on comparing CS against AD (or reference) in the testing.

tests/reg_tests/test_jacvecprodFWD.py
joanibal: > Then you would re-train at that point.

Yeah, you could re-train. It just bothers me that AD and CS could agree but still be flagged as incorrect because they are compared using old values. 

> Yeah it's doable but I don't get why that's useful here. doing CS over AD kinda gives you second derivatives. Plus its much more pain than what you are trying to do here.

I don't want to get second derivatives, I just want to run both the master_d/b without an imaginary perturbation and the master routine with an imaginary perturbation.
That way I can have the AD and CS values in memory at the same time and I can compare them without writing either out to a file first. You can do this in TACS for example. 

With that being said. I don't think this is important enough to implement now, so I'm just going to go the reference file route.

joanibal:
@anilyil 
I have added new tests to compare CS with AD. 
I also fixed a bug I found because of this new test....

anilyil:
Sorry for the delay. The changes look good to me. Before merging, I would like to update the PR description though. The title and the description only mentions the added functionality, but the PR also has 3 bug fixes or changes not directly related to this:
1.  The vec get/restore stuff for `xSurfVec` in the master routine.
2. Changing the order of calls with `timestep_block` in the master routines.
3. The `alwaysRealType` check in the pyf preprocessor, which addresses part of #133.

I can also edit the PR description, but can you please remind me why we changed the order of `timestep_block` calls, @joanibal?  I know we discussed in person or within some of the resolved discussions here but I think it will be worth putting this info up in the description to avoid people needing to dig into the PRs to figure out what we were doing if this change breaks something down the line.

joanibal:
I moved the call of `timestep_block` to be consistent within each of the master routines. Before there were slight differences in the call order in each routine.

joanibal:
The description has been updated. Let me what you think @anilyil?

anilyil (approved):
Thanks for all the modifications!

joanibal:
@SichengHe Have I addressed all of your comments?

SichengHe:
It looks good!

SichengHe (approved):

PR #176, opened 2021-12-15
Improved cavitation constraint

yqliaohk:
## Purpose
This PR includes changes to cavitation constraint. I modified and reorganized it so it is backward-compatible with the original form while also can be used for the formulation I used in my thesis.

## Type of change

## Testing
- Backward-compatibility test: I first retrained the `test_cl_solve.py:TestSolve` test without updated code to generate the new ref file that include cavitation as a target function (with computecavitation True and cavitation added as one of evalfuncs). Then I used the updated code to run the reg test. The reg test passed.
- Complex step test 1: I used the default options with updated code to retrain the ref file `adjoint_rans_tut_wing.json` (with computecavitation set to True and cavitation added as one of evalfuncs), and then ran complex step test. The test passed.
- Complex step test 2: Same test as the previous one, but with `cavExponent` set to 2, which is the formulation I used in my thesis.
- I reran one of my previous hydro-only case and the evaluated cavitation function value matched 7 digits with my previous result. 

## Checklist


gang525:
Just so we're on the same page and it is documented here since it came up in the maintenance meeting, @yqliaohk what are the remaining action items to do before this branch can be merged?
I believe the last time we spoke, you said you want some regression tests added. Maybe it would also be good for me to have a completed optimization using the cavitation, though my derivatives are all agreeing with the FD approximations from SNOPT.

yqliaohk:
> Just so we're on the same page and it is documented here since it came up in the maintenance meeting, @yqliaohk what are the remaining action items to do before this branch can be merged? I believe the last time we spoke, you said you want some regression tests added. Maybe it would also be good for me to have a completed optimization using the cavitation, though my derivatives are all agreeing with the FD approximations from SNOPT.

No action items remaining now. Once I clean the ref files, it will be ready to merge.

gang525 (approved):
The implementation works as expected

nwu63:
I personally think we should merge stuff instead of letting stuff sit stale since I assume further development of this will not be anyone's priority. @anilyil  what do you think?

gang525:
I talked with Anil about a KS-based cavitation constraint and its definitely something we'll work on getting merged when its ready. It can potentially be a second option for the cavitation constraint if we decide their are specific pros and cons of each method for constraining cavitation. Right now though, i also agree we should just try to merge yingqians code.

src/inputParam/inputParamRoutines.F90
```diff
@@ -4042,6 +4042,9 @@ subroutine setDefaultValues
     usematrixfreedrdw = .False.
     sepSensorOffset = zero
     sepSensorSharpness = 10_realType
+    cavSensorOffset = zero
+    cavSensorSharpness = 10_realType
+    cavExponent = 0
```
sseraj, line 4047: I'm pretty sure these lines are not necessary because the options get set in Python anyway

src/output/outputMod.F90
```diff
@@ -2206,7 +2206,7 @@ subroutine storeSurfsolInBuffer(sps, buffer, nn, blockID,   &
              plocal = half*(pp1(i,j) + pp2(i,j))
 
              sensor1 = (-(fact)*(plocal-pInf))- cavitationnumber
-             sensor1 = one/(one + exp(-2*10*sensor1))
+             sensor1 = (sensor1**cavExponent)/(one + exp(-2*cavSensorSharpness*(sensor1 - cavSensorOffset)))
```
sseraj, line 2209: I think it would be better to have the signs match the papers/thesis
`sensor1 = (sensor1**cavExponent)/(one + exp(2*cavSensorSharpness*(-sensor1 + cavSensorOffset)))`

src/solver/surfaceIntegrations.F90
```diff
@@ -505,7 +505,7 @@ subroutine wallIntegrationFace(localValues, mm)
           tmp = two/(gammaInf*MachCoef*MachCoef)
           Cp = tmp*(plocal-pinf)
           Sensor1 = -Cp - cavitationnumber
-          Sensor1 = one/(one+exp(-2*10*Sensor1))
+          Sensor1 = (Sensor1**cavExponent)/(one+exp(-2*cavSensorSharpness*(Sensor1-cavSensorOffset)))
```
sseraj, line 508: Same comment here about the signs

sseraj (commented):
> Once I clean the ref files, it will be ready to merge.

What's the status on this? It looks like there was a lot of effort put into testing, so it would be nice to have that added to the repo.

Also, the docs build is failing because each new option needs a description in options.yaml. This would be the place to put things like this
> cavExponent set to 2, which is the formulation I used in my thesis

yqliaohk:
> > Once I clean the ref files, it will be ready to merge.
> 
> What's the status on this? It looks like there was a lot of effort put into testing, so it would be nice to have that added to the repo.
> 
> Also, the docs build is failing because each new option needs a description in options.yaml. This would be the place to put things like this
> 
> > cavExponent set to 2, which is the formulation I used in my thesis

@sseraj Were you talking about adding all the testing I did in the description to the repo?

src/inputParam/inputParamRoutines.F90
yqliaohk: Is it better to not have them? I followed the format for `sepSensorOffset` etc

sseraj:
> @sseraj Were you talking about adding all the testing I did in the description to the repo?

No, not all of it, just whatever you had mind.

src/inputParam/inputParamRoutines.F90
sseraj: I think it's better to remove them. Right now, there are "default" values in three different places in the code and leaving them in there makes it unclear which ones actually do anything.

sseraj (approved):

PR #180, opened 2022-01-29
Changed printing to use proper scientific notation

nwu63:
## Purpose
I changed a lot of the printing that used the format `e.` to use `es.`, so that numbers show up as `4.9E2` instead of `0.49E3`. I was really annoyed with the constant leading zero, so this should make everything more readable.

I used some regex to find these, so I might've missed some stuff. Would appreciate some checking by ADflow maintainers since I'm not as familiar with the code. This may also be a good time to change other printing formats, especially since I noticed we are now showing 16 decimal places which is probably too much.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I ran the MACH-Aero tutorial and saw that the printing is now using proper scientific notation.

## Checklist

### DIFF
--- a/src/adjoint/adjointAPI.F90
+++ b/src/adjoint/adjointAPI.F90
@@ -873,7 +873,7 @@ contains
 
 10  format(a)
 20  format(a,1x,f8.2)
-30  format(1x,a,1x,e10.4,4x,a,1x,i4)
+30  format(1x,a,1x,es10.4,4x,a,1x,i4)
 40  format(1x,a,1x,i5,1x,a)
 
 #endif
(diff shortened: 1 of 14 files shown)


eirikurj:
@nwu63 I updated one more place that I was aware was missing. 
Perhaps we should also consider updating how the slice/lift .dat files write to the files, as they also use the leading zero notation.

nwu63:
I agree, I've pushed up some more changes. Please take a look.

sseraj (approved):
I tried this out locally and it is a definite improvement. I actually like 16 digits being printed for the monitor variables, so I am fine with this as is. I also checked that the Tecplot output files still work with the new format.

anilyil (approved):

Issue #183, opened 2022-01-30
segfault when writing cgns surface solutions that include  yplus in the surfacevariables list on pleiades.


anilyil:
## Description
The adflow build on the new operating system on pleiades segfaults when writing the cgns surface output file, but only when the variable `yplus` is included in the `surfacevariables` option. I can only recreate this issue with the podded fan overset mesh, and it does not show up with the mach-tutorial wing case. 

With the aeroprop case, the issue is there even on a single node, including a 1 proc run and a saturated node run. Its also there with a 4 node job with saturated nodes. 

One day before this, the tecplot surface output was also broken but it mysteriously fixed itself after a re-make of adflow. 

### Steps to reproduce issue

1. On pleiades, run any of the aeroprop benchmark cases with yplus in the surface variables list and write a cgns surface output.

### Current behavior
segfault

### Expected behavior
no segfault

### Code versions
- Operating System: TOSS 3
- Python: 3.9.5
- CGNS: 4.2.0 (also there with 4.1.2)
- PETSc: 3.15.4
- This software: https://github.com/mdolab/adflow/commit/d17ca9ffc33392c17533ed6d83dd056962b1c741
```
Currently Loaded Modulefiles:
  1) pkgsrc/2021Q2           2) comp-intel/2020.4.304   3) mpi-intel/2019.5.281    4) gcc/9.3                 5) python3/3.9.5
```

nwu63:
_probably_ a duplicate of #181, since `yplus` is in the surface variables there also. I also could not reproduce this with MACH-Aero, it came up in a relatively large multiblock mesh for me.

anilyil:
I am not getting the same runtime error but yeah they could be the same issue. Can you try without the yplus output?

nwu63:
I can confirm that without `yplus` I do not get the segfault.

anilyil:
definitely sounds like a duplicate of #181. Closing this since your issue has more detailed info

Issue #181, opened 2022-01-29
Segfault due to #158


nwu63:
## Description
I started getting segfaults recently in an aerostructural optimization script, and I used `git bisect` to trace it to #158. Unfortunately, I'm not able to reproduce this issue in the simple analysis case in MACH-Aero, but this is the log:

```
At line 1923 of file ../output/outputMod.F90
Fortran runtime error: Index '1' of dimension 1 of array 'viscpointer' below lower bound of 2
```
which corresponds to the line
```
mm = viscPointer(ii,jj)
```

Not entirely sure what the call graph here is, but it's happening at some point when writing the surface CGNS solution.

### Steps to reproduce issue
Haven't been able to reproduce this, but can help run cases and debug.

### Current behavior
There's a segfault

### Expected behavior
There shouldn't be a segfault

### Code versions

- Operating System: 
- Python: 
- OpenMPI: 
- CGNS: 
- PETSc: 
- This software: master branch

joanibal:
Hmm, are you able to share the mesh you used?

Could you post your `surfacevariables`?

nwu63:
```
"surfaceVariables": ["cp", "vx", "vy", "vz", "yplus", "sepsensor"]
"isosurface": {"shock": 1.0, "vx": -0.001},
```
Let's talk offline regarding the mesh.

PR #185, opened 2022-01-31
Fixed surface i,j extents in output to prevent segfaults

joanibal:
## Purpose
Addresses #181 #183 

The iBeg, iEnd and jBeg, jEnd where actually of the zone and not the subface as I thought. 
I have corrected this and changed the variable names to help avoid further confusion. 

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Run cases that were previously segfaulting @nwu63 and @anilyil.

## Checklist

### DIFF
--- a/src/output/writeCGNSSurface.F90
+++ b/src/output/writeCGNSSurface.F90
@@ -387,7 +387,11 @@ contains
 
     integer(kind=intType) :: i, offset
     integer(kind=intType) :: mm, mBlocks, faceID, nSubfaces
-    integer(kind=intType) :: iBeg, jBeg, kBeg, iEnd, jEnd, kEnd
+    
+    ! extents of the subface in terms of the i,j,k indices of the cgns zone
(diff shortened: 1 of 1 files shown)


joanibal:
I suspect that is wasn't a problem for most since iBeg, iEnd and jBeg, jEnd where of the zone are the same for the subface when k is the off wall direction. This is the case with pyHyp meshes.

nwu63 (approved):
I don't know the details of the changes, but I can confirm that the fix works for my case.

anilyil (changes requested):
My case still fails when I have `yplus` in the surface output list. Because `cf` and `cfx` etc. are also handled the same way (https://github.com/joanibal/adflow/blob/visc_surf_out_fix/src/output/outputMod.F90#L1872), they also segfault.

anilyil (approved):

nwu63 (approved):

PR #188, opened 2022-02-10
Implemented restartAdjoint option

sseraj:
## Purpose

I implemented the restartAdjoint option, which previously did nothing. The default behavior, which is to use the previous adjoint solution as the initial guess, is unchanged.

Using `"restartAdjoint": False` improved my optimization convergence slightly. I don't know if this will be helpful in general, but it's a nice option to have even if it is just for debugging.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -3405,11 +3405,16 @@ class ADFLOW(AeroSolver):
         # on this AP, there is no point in solving the reset, so continue
         # with psi set as zero
         if not (self.curAP.adjointFailed and self.getOption("skipafterfailedadjoint")):
-            # Extract the psi:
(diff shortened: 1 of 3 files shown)


marcomangano (approved):
Nice! Did you notice significant increases in computational wall time?

sseraj:
> Nice! Did you notice significant increases in computational wall time?

For my case, there was no increase in wall time for the adjoint solution. This is because the adjoint convergence stalls before reaching the prescribed tolerance (due to #4), so the solver always runs for the maximum number of iterations. For a regular case where the adjoint converges normally, I would expect a slight increase in wall time.

anilyil (approved):

PR #193, opened 2022-02-23
Zipper mesh segfault fix

sseraj:
## Purpose

This fixes the zipper mesh segfault issue first reported by @awccopp. 

When creating the zipper mesh, indexing out of bounds can occur when the number of closest points we are searching for is greater than the total number of points on a string. Capping the initial `nSearch` in the `stringMatch` routine solves the issue.

The current initial value for `nSearch` is 50, which explains why the problem only appears on small meshes like @awccopp's airfoil cases. I added a test with a simple overset cube mesh that can reproduce the error.

## Expected time until merged

Regular urgency; one week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Run test_zipper.py without compiling the fix. The result will be a segfault (or 'index outside of expected range' if compiled with a bounds check flag).
Run the test again after recompiling. The test should now pass.

## Checklist

### DIFF
--- a/src/overset/stringOps.F90
+++ b/src/overset/stringOps.F90
@@ -2571,8 +2571,9 @@ module stringOps
        ! Loop over my nodes and search for it in master tree
        nodeLoop:do j=1, str%nNodes
 
-          ! Reinitialize initial maximum number of neighbours
-          nSearch = 50
+          ! Set the initial maximum number of neighbours
+          ! This can be at most the total number of nodes
+          nSearch = min(nAlloc, master%nNodes)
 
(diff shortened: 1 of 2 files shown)


joanibal:
Nice work Sabet!

anilyil (approved):

anilyil:
it looks like we need another approval? I tried requesting one from @joanibal since he already approved but github didnt like that

awccopp (approved):
I pulled your fixes and ran them locally on some cases where I had this issue and they all ran fine. Looks good to me thanks for fixing.

PR #196, opened 2022-03-11
adding new mphys wrapper and removing the old openmdao wrappers

anilyil:
## Purpose

This PR removes the old `om_adflow` wrapper and moves the new MPhys ADflow wrapper from the MPhys repo to the ADflow repo. Importing ADflow on its own should not trigger an import of the MPhys wrapper, so we also do not need the OpenMDAO warning. To import the new mphys wrapper:
```
from adflow.mphys import ADflowBuilder
```
which is the only class needed from ADflow for MPhys.

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
We can discuss how we want to add tests now that MPhys is open-source and the solver repos have their own builders.


anilyil:
What do you recommend for testing @joanibal? Do we want to just move what we had in MPhys or do we want to add new tests on top?

joanibal:
I'd like to add some tests here too. With all the changes we make to ADflow, It's worth checking that they didn't break the MPhys wrapper.

joanibal:
Should we copy the tests? or maybe we can run the ADflow tests in the mphys repo when changes are made to ADflow.

anilyil:
I would like to have a few adflow only mphys tests here. not sure how we can only run adflow tests on mphys side when adflow changes

sseraj (commented):
Please also fix the style checks

joanibal (changes requested):
Unfortunately, I could not get the mphys wrapper to import properly when not using `pip install -e .`.
I tried `pip install .`, but it didn't work for me.

anilyil:
Right. Is there a better approach for the import here? I can check if openmdao is available and then import but that results in weird side-effects

joanibal:
It appears to still be broken

```
mdolabuser@josh-mdolab:~/repos/mphys/tests/regression_tests$ python test_aerostruct.py 

pyOCSM version 1.19 has been loaded

Traceback (most recent call last):
  File "test_aerostruct.py", line 31, in <module>
    from adflow.mphys import ADflowBuilder
ModuleNotFoundError: No module named 'adflow.mphys'
```

nwu63:
Did you try either of the approaches I mentioned? I don't think Anil pushed anything related to setup.py.

nwu63:
@joanibal I pushed up the changes and they work locally for me.

anilyil:
I tested @nwu63's changes and they worked great for me. I dont think it imports the mphys builder by default, it can import it regardless of an in place install or a site packages install, and it complains that I dont have openmdao if I try to import it w/o having openmdao installed, but I can still import the ADFLOW class. 

On top of this, I added the requirements list for the mphys wrapper because it uses openmdao, mphys, and also idwarp. 

@joanibal can you check the imports again with the current state of the PR?

joanibal:
It works for me after the changes

joanibal (approved):

sseraj (approved):

PR #203, opened 2022-03-18
GMRES orthogonalization fix

### LINKED ISSUE #4: Zipper mesh bug
Fix the bug causing the gradients to be off when using zipper meshes.

### LINKED ISSUE #195: Unstable orthogonalization in GMRES
## Description

By default, PETSc's GMRES uses the classical (unstable) Gram-Schmidt method for orthogonalization, apparently because it is faster than the modified (stable) Gram-Schmidt method. (See the documentation [here](https://petsc.org/main/docs/manualpages/KSP/KSPGMRES.html).) This leads to instabilities when using large subspace sizes, which is especially problematic for the adjoint solution. Instabilities in the adjoint solution can result in the KSP Residual not matching the actual residual (the 'Norm of error' computed in ADflow), spurious convergence, and inaccurate gradients.

The fix is simple and I have implemented it [here](https://github.com/sseraj/adflow/commit/008780f8af062cc9aa7ed1208b87ed78787da3ec).

I have not found any slowdown when using the stable method. The time to run all the tests with `testflo -v` on 16 cores was similar.
```
Unstable: 00:13:52.15
Stable: 00:13:23.89
```

I also tested the runtime on a large scale adjoint case and the stable method was faster. Running one adjoint solution for a 1.4M cell overset mesh with 300 adjoint iterations and a subspace size of 300 on 16 cores took:
```
Unstable: 843.928 sec
Stable: 679.646 sec
```

I see three possible ways of implementing this fix (in my order of preference):

1) Always use the stable orthogonalization as I have currently implemented. 
2) Add an option to use the stable orthogonalization, defaulting to True.
3) Add an option to use the stable orthogonalization, defaulting to False. This would preserve backwards compatibility, but I see no reason to do this.

I would like to use this issue to discuss which of these to implement.

sseraj:
## Purpose

This changes the default orthogonalization in GMRES to the modified Gram-Schmidt method and closes #195. The issue has more details on the performance and accuracy of the different orthogonalization methods. 

I also made some minor changes related to the adjoint:
- Uncommented some timing code, which fixed the adjoint time always being zero
- Added a note in the docs on always using the matrix-free adjoint with overset
- Added a note in the docs about looking at the absolute adjoint residual if the relative convergence stalls

## Expected time until merged

One week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


sseraj:
The overset adjoint test is failing. I'm not sure why, but it's not related to the orthogonalization fix because all the other tests pass. To avoid delaying the fix, I will remove the test and try adding it again some other time.

doc/options.yaml
```diff
@@ -1119,6 +1121,15 @@ adjointSolver:
   BCGS: BiCGStab
   IBCGS: Improved BiCGStab
 
+GMRESOrthogonalizationType:
+  desc: >
+    Which orthogonalization method to use with GMRES.
+    This will have the most effect on the adjoint solution, but the same method is also used in the ANK and NK solvers.
```
anilyil, line 1127: Can you rephrase it slightly? The first time I read it I thought you were talking about the option's effect on the final adjoint state rather than the linear solver algorithm. The adjoint solution is the final linear state, adjoint solution process is the actual process, or you can call it the "adjoint solver". 

It is also true that it does affect the final solution. So maybe something like:
" This will affect the adjoint solver the most, but the same option is also used in the linear solvers in the ANK and NK solvers." in the main description, and then you can clarify a bit more about what you mean by "accuracy" in the explanation of each option below.

doc/options.yaml
```diff
@@ -1202,8 +1213,9 @@ useMatrixFreedrdw:
   desc: >
     Flag that determines if ADflow will use the matrix-free routines to compute mat-vec products with the full Jacobian.
     If False, ADflow will form the full matrix exactly and mat-vec multiplications will be matrix-based.
+    This should always be True for overset meshes, otherwise the gradients will be inaccurate.
```
anilyil, line 1216: maybe mention this is an unresolved bug and link to an issue on github? There's no inherent reason why a matrix-based adjoint should not work for overset meshes.

src/adjoint/adjointAPI.F90
```diff
@@ -815,8 +815,7 @@ subroutine solveAdjoint(RHS, psi, checkSolution, nState)
        ! Determine the maximum time using MPI reduce
        ! with operation mpi_max.
 
-       ! call mpi_reduce(timeAdjLocal, timeAdj, 1, adflow_real, &
-       !      mpi_max, 0, ADFLOW_COMM_WORLD, ierr)
+       call mpi_reduce(timeAdjLocal, timeAdj, 1, adflow_real, mpi_max, 0, ADFLOW_COMM_WORLD, ierr)
```
anilyil, line 818: I wonder why this call was removed. I know Sandy was doing some timing studies during the ADflow paper, maybe happened around then?

anilyil (changes requested):
Excellent find and fix. Thank you @sseraj for working on this. I left a few comments but they are all about the explanation of the option and comments.

src/adjoint/adjointAPI.F90
sseraj: It was commented out in [this commit](https://github.com/mdolab/adflow/commit/aa69a61ee47dbb2d19f312f6f54904b1fd1dfd29), probably just for debugging and then left like that by mistake.

anilyil (approved):

PR #218, opened 2022-05-09
Increased ANKSwitchTol

sseraj:
## Purpose

As discussed in the maintenance meeting, I increased `ANKSwitchTol` such that we are practically guaranteed to start with ANK for all cases. A switch tolerance of 1.0 will still use the smoothers when design changes increase the residual above the reference, or even on the first call in some cases as @nwu63 found (probably due to floating point precision).

## Expected time until merged

1-2 days

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/reg_tests/test_solve.py
+++ b/tests/reg_tests/test_solve.py
@@ -124,6 +124,9 @@ class TestSolve(reg_test_classes.RegTest):
 
 
 class TestFailedMesh(unittest.TestCase):
+
+    N_PROCS = 1
+
     def test_failed_mesh(self):
         options = copy.copy(adflowDefOpts)
         options.update(


tests/reg_tests/test_solve.py
```diff
@@ -124,6 +124,9 @@ def test_solve(self):
 
 
 class TestFailedMesh(unittest.TestCase):
+
+    N_PROCS = 1
```
nwu63, line 128: Just FYI that this is already the default in testflo, but I guess this is more explicit.

nwu63 (approved):

tests/reg_tests/test_solve.py
sseraj: I think `N_PROCS = 1` will run with MPI but with one processor. The test in the terminal has `(mpi)` before the path and no other terminal output is printed. I made this change to avoid having the ADflow header printed out during the tests.

ArshSaja (approved):

PR #228, opened 2022-08-03
Compile with `-std=f2008`

nwu63:
## Purpose

This the last remaining repo. This fixes a number of things
- compilation with strict `-std=f2008` standard
- compilation under GCC>= 10

I got the real build to work (there are some warnings still but it should run fine). The complex build has some extra problems that will require more investigation, in particular `complexify.py` probably needs to be modified. I have absolutely no idea how it works in idwarp without such fixes.

I will not continue working on this PR, but this should be done soon as it is blocking adoption of our codes with newer GCC compilers.

## Expected time until merged

This is not ready to merge yet.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


eirikurj:
@nwu63 I can have a look at the remaining issues as I wanted to update the complex built anyway.

config/defaults/config.LINUX_INTEL.mk
```diff
@@ -27,7 +27,8 @@ CGNS_INCLUDE_FLAGS=-I$(CGNS_HOME)/include
 CGNS_LINKER_FLAGS=-L$(CGNS_HOME)/lib -lcgns
 
 # ------- Define Compiler Flags ----------------------------------------
-FF90_FLAGS   = -DHAS_ISNAN -fPIC -r8 -O2 -g
+FF77_FLAGS   = -DHAS_ISNAN -fPIC -r8 -O2 -g
+FF90_FLAGS = $(FF77_FLAGS) -std=f2008
```
sseraj, line 31: Doesn't Intel have a different standard flag?

eirikurj: Yes, they do. Should be updated now.

nwu63:
Is this ready @eirikurj?

eirikurj:
Yeah, I think this is good for now. Will mark it ready for review. Few things to note. I created an interface for `myisnan` to handle both the real and complex checks. C function `myisnan` is removed in favor of a more modern fortran module. Previously there was not really a check for complex numbers, but here I decided just to check both real and imag parts. This is thus a change in behavior. Also, despite the name `myisnan` the C function also checked for `INF` so I retained this behavior in this implementation.  Please make sure to test this thoroughly in case you have problems that you can trigger `NAN` or `INF` for complex case.

src/output/writeCGNSSurface.F90
```diff
@@ -186,7 +186,7 @@ subroutine writeCGNSSurfaceSol(famList)
              if (ierr /= CG_OK) &
                   call terminate("WriteCGNSSurfaceSol", &
                   "Something wrong when calling cg_base_write_f for &
-                  isoSurface")
+                  &isoSurface")
```
sseraj, line 189: Some of these changes would be better if we just removed the line break. I can change these in a future PR though

sseraj (dismissed):
Looks good. I think this is fine as is so I approved, but I also have some comments:

I have a failing case that produces NaNs in complex mode. The behavior did change with this PR. Instead of crashing with returnFail during the first iteration, it continuously takes NK steps of size 0.01 where all monitor variables are NaNs. I prefer the original behavior, but this is not that big of an issue.

eirikurj:
@sseraj this is a bit much of a regression for my liking. The `returnFail` should handle this. Do you have a MWE that I could check.

src/output/writeCGNSSurface.F90
eirikurj: Yeah, there are more of these. I can update this one here, as this has already been touched.

sseraj:
Unfortunately, I do not have a MWE. I did some more testing and found that the change in behavior is because `ieee_arithmetic` does not work with the `-ffast-math` flag, which is on our default gfortran config. As a result, the following check does not work.

https://github.com/mdolab/adflow/blob/537ae3c31105193fc1707c1946d13241844c0198/src/solver/solvers.F90#L1680

Removing `-ffast-math` results in the same behavior as before. This might also be an issue in real mode.

nwu63:
Oh I didn't realize we are doing `-ffast-math` by default, that doesn't seem right. I don't think we're doing that in other packages right?

eirikurj:
Ahh, of course, I forgot about this flag. This makes sense. Any check for NaN will not work with this flag, as enabling will turn on `-funsafe-math-optimizations, -ffinite-math-only`, so NaNs effectively do not exist in the code and are optimized out. I see few options:
- Compile this module without the `-ffast-math` optimization flag. We could move `myisnan` to another module and apply no optimization flags.
- Not compile our code with `-ffast-math`.
- Use the older c function, as that is not compiled with this flag.
I would prefer the first or second option, at a minimum using a fortran module with all the benefits (as opposed to C interface with no type checking) and preferably avoid custom isnan solutions as they might not be portable.
@nwu63 yeah dont think so either. Its somewhat questionable using this flag in this type of code as it might introduce inaccuracies, changing the results slightly. Its been a while since I looked at performance with and without the flag (and inaccuracies for that matter), but perhaps we should not be using this flag as a default flag for gfortran. I think "apply at own risk" is probably more appropriate.  Let me know what you guys think.

sseraj:
I ran the tests with `testflo -v` on 16 cores and got these times:
With `-ffast-math`: 16:56.60
Without `-ffast-math`: 17:59.36

We don't get too much speedup with `-ffast-math`, so I would be okay removing it from the default config. If we really want to keep the performance, moving `myisnan` to another module makes sense.

anilyil:
I would say at least in the config we use for the clusters we should try to include whatever optimizations are possible. The tests are not a great estimate of performance, but still, the difference is there right

sseraj:
I agree the cluster builds should include optimizations. Do we still use gfortran on Great Lakes? If so, maybe compiling `myisnan` without the flag is better than removing it entirely.

nwu63:
> I agree the cluster builds should include optimizations. Do we still use gfortran on Great Lakes? If so, maybe compiling `myisnan` without the flag is better than removing it entirely.

Yes we do. Maybe we should discuss more broadly the use of `-O3` or `-Ofast`/`-ffast-math` in production runs vs. the more conservative `-O2`. Also [this](https://simonbyrne.github.io/notes/fastmath/) may be helpful to read as a reference. I don't know if we're okay with non-IEEE compliant compiler flags in general and I think it's hard to quantify the benefits/risks in a comprehensive manner.

eirikurj:
@nwu63 sure we have a broader discussion on this, especially for production runs.  

I am somewhat mixed on this. Using `-O3` should probably be fine, but I am more worried about `-Ofast` or `-ffast-math`. Here we can really run into some unexpected behavior without realizing. I think the default flags should be something that is reasonable, safe, and will give expected behavior. So perhaps a `-O2` is more appropriate as the default.

In principle, we should check the output for each optimization to make sure nothing is largely off. For fun, I ran the MACH-Aero tutorial locally with 8 cores with different compiler flags. 

  | Time (avg) | Relative  | wing_cd | wing_cl
-- | -- | -- | -- | --
No opt | 121.19 | 220.7 | 0.016801751517189 | 0.417763635022066
-O | 47.34 | 25.3 | 0.016801751224581 | 0.417763634353593
-O2 | 42.02 | 11.2 | 0.016801751322668 | 0.417763629144478
-O3 | 41.12 | 8.8 | 0.016801751327773 | 0.417763634965067
-ffast-math -O3 | 37.79 | 0 | 0.016801751309968 | 0.417763634912151

Based on this (very simple test), there is benefit to be had by using these flags, without altering the function values much (relatively speaking). The issue is, as we see with complex values, that the behavior is different. If we want to maintain the same behavior, we need to do compile the modules separately.

nwu63:
Ideally, we generate training data without any IEEE-unsafe optimizations, and we run tests with whatever flags we would be running on production cases. I'm not sure how feasible that is though. An alternative is to just not have any unsafe flags in default, and tell people who are compiling on clusters to use them at their own risks.

sseraj:
Regardless of what we choose as the default, being able to use `-ffast-math` and also catch NaNs as we currently do is pretty useful. Can we still do this by compiling `myisnan` separately?

eirikurj:
Right, yes, I would prefer a consistent behavior if possible. Alright, I will make some changes.

eirikurj:
I pushed a proposal to allow `-ffast-math` flag to be used without altering NaN behavior to a separate branch and made a PR here https://github.com/mdolab/adflow/pull/233 @sseraj can you use your complex case to test?

sseraj:
> @sseraj can you use your complex case to test?

Yes, I will test this.

sseraj:
I opened issue #234 to keep track of updating the opt flags based on what we discussed in the maintenance meeting. I think it makes sense to update the flags and retrain the tests in another PR.

sseraj (approved):

anilyil:
I did want to take a look at this. I can review within the next 2 days if there is no urgency. I wanted to review partly also because my current open PR touches a bit of fortran and I want to see how bad the conflicts will be and if I can fix that PR as well.

anilyil (commented):
Most of the changes here look good. I have a few general comments:
1. None of my cases get NaNs right now, so I cannot test this. However, if you want, we can add a test that we force to get a nan and see if we can catch it. 
2. What did you mean by try my "complex" cases? As in complicated cases or cases that use complex numbers for complex step checks? 
3. Some files only have whitespace added (adBuffer.f, adStack.c). Do we want to keep these whitespace changes or just remove them to have a cleaner diff? 
4. Will this now not work with compilation using GCC 9? 

Otherwise stuff look good. I don't fully understand the implications of the compilation flags but it would be nice if one of you can give me a quick summary over a call so I dont miss any nuance. The other code changes make sense.

joanibal:
Am I correct that this PR is necessary but not sufficient to get ADflow working with GCC 11?

joanibal:
With a cheeky little `-fallow-argument-mismatch` I'm able to run it with gcc 11

joanibal (approved):
Works for me.

nwu63:
> With a cheeky little `-fallow-argument-mismatch` I'm able to run it with gcc 11

I would like to eventually fix these, so that we do not have to conditionally add those to the Docker builds once we update to U22 which will be using gcc>9.  But I guess that does not have to happen in this PR

joanibal (approved):

sseraj:
@anilyil @eirikurj Any updates on this? We should probably merge this soon.

eirikurj:
@anilyil wanted to have a chat offline, but we have yet to meet. Will reach out offline so we can figure out a time and can get this merged soon.

eirikurj:
@sseraj, @anily plans to review this later this week.

anilyil (approved):

sseraj (approved):

PR #235, opened 2022-10-05
raise Error if a Mesh Cell is skewed above threshold

DavidAnderegg:
## Purpose
I am optimising a two element-airfoil. Because of the proximity of those elements, the mesh quality deteriorates fast. To fix this, i check if the variable `adflow.killsignals.fatalfail` is set to `true`. When this is the case, I regenerate a mesh.
Unfortunately, this method is to slow as the variable is only set when the mesh has negative volumes. 

This PR should addresses this in the following way: 
Calculate the skewness as described [here](https://www.simscale.com/docs/simulation-setup/meshing/mesh-quality/#skewness). If it goes above a certain threshold defined in the regular python-wrapper (e.g. `meshMaxSkewness: 0.4`), ADflow errors out by means of setting `adflow.killsignals.fatalfail` to `true`.

I do not consider this PR ready yet. But I want to get feeback from you guys as soon as possible. I am especially interested in these points:

1. I only calculate the skewness when we are actually interested in it (`meshMaxSkewness` < 1). Does this make sense?
1. Is highjacking the subroutine `WriteNegVolumes` fine as I did?
1.  I need to release my new `flowdom` variable `skew`. Where is the best place to do this?

## Expected time until merged
Probably a couple of weeks.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

tbd

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -4968,6 +4968,7 @@ class ADFLOW(AeroSolver):
             ],
             "turbulenceOrder": [str, ["first order", "second order"]],
             "turbResScale": [(float, list, type(None)), None],
+            "meshMaxSkewness": [float, 1.0],
             "turbulenceProduction": [str, ["strain", "vorticity", "Kato-Launder"]],
             "useQCR": [bool, False],
             "useRotationSA": [bool, False],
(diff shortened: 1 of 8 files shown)


eirikurj:
Changing to draft as author suggest that the PR is not ready and content needs discussion.

PR #240, opened 2022-10-22
Eval funcs docs

gang525:
## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


gang525:
@ArshSaja Please document the remaining functions. To others, feel free to check the changes I have made so far so we can iterate. Thanks!

gang525:
Yep, this is exactly the place to put that documentation. It is not currently in this PR so I will add it to my TODO to specify that the compute cavitation flag results in extra computations in the cost function description.

ArshSaja:
I will work on this during the break.

ArshSaja:
This PR is ready for the review..

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+  desc: >
+    Averaged mass flow in the y direction.
+
+mavgvz:
+  desc: >
+    Averaged mass flow in the z direction.
+
+cperror2:
+  desc: >
+    The square of the different betwween computed cp and target cp*.
```
nwu63, line 405: _difference_

We should probably at a minimum link to the function `setTargetCp`. This I believe is for performing inverse design where we want to match a given cp distribution. This would be the L2 norm of the error which we can directly feed into the optimizer as the objective.

nwu63 (commented):
Left some minor comments below. Keep in mind I'm not familiar with a lot of the stuff in there (actuator disk, TS, etc.) so I would suggest finding others to review this also.

How similar is the sphinx extension to the one we already have? I wonder if we want to do something a bit more sophisticated (subclass one from the other, or create a base class or something) just to reduce the code duplication. Also if we think this would be used anywhere else in the future then we may want to throw it into the sphinx package, although at the moment I think this might be the only repo.

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+mdot:
+  desc: >
+    Mass flow rate in the freestream direction.
+    This quantity is important for computing momentum forces.
+    Units: kg - s:math:`^{-1}`
+
+mavgptot:
+  desc: >
+    Mass flow rate averaged total pressure 
+    # TODO: GN- I have no idea if this is right based on the source code --> surfaceIntegrations.F90 --> search for 'mass_Ptot'
```
joanibal, line 289: Yes. This is right. You can remove the TODO

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+aavgps:
+  desc: >
+    Cell area averaged static pressure
+    Units: Pascal
+
+mavgttot:
+  desc: >
+    Mass flow rate averaged total temperature 
+    # TODO: GN- I have no idea if this is right based on the source code --> surfaceIntegrations.F90 --> search for 'mass_Ttot'
```
joanibal, line 305: Yep, this is right.

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+  desc: >
+    Pressure component of lift
+
+liftviscous:
+  desc: >
+    Viscous lift
+
+liftmomentum:
+  desc: >
+    Momentum lift (due to changing momentum of flow?)
```
joanibal, line 389: I'm also confused here. I'm not sure what this means.

joanibal (approved):
Looks good to me. I have only a few comments.

doc/conf.py
```diff
@@ -23,16 +26,13 @@
...
 # mock import for autodoc
-autodoc_mock_imports = ["numpy", "mpi4py", "petsc4py", "baseclasses", "adflow.om_adflow"]
+autodoc_mock_imports = ["numpy", "mpi4py", "petsc4py", "baseclasses"]
 
 # bibtex sources
 bibtex_bibfiles.extend(["citations.bib"])
 
-# intersphinx
-intersphinx_mapping = {
-    "mach-aero": ("https://mdolab-mach-aero.readthedocs-hosted.com/en/latest", None),
```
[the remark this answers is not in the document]
sseraj: `intersphinx_mapping` is defined in sphinx-mdolab-theme, so this is unnecessary.

As for the Sphinx extension, enough of the extension is specific to pyADflow that it will be difficult to generalize nicely.

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+  desc: >
+    NOTE: Time spectral stability derivatives are broken as of 2022.
+    TODO: Someone who knows aircraft stability should verify these
+
+clqdot:
+  desc: >
+    NOTE: Time spectral stability derivatives are broken as of 2022.
+    TODO: Someone who knows aircraft stability should verify these
+
+cbend:
```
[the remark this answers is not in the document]
marcomangano: Nope, I don't know what this refers to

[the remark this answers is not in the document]
nwu63: Isn't this the bending moment we use in lieu of adding structural modeling? I thought you used this in the supersonic work before.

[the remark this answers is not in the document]
sseraj: This functional is described here: https://github.com/mdolab/adflow/blob/d5fe4623cd825aa1e523cdffb844614bbb72db4d/src/solver/surfaceIntegrations.F90#L178-L180

It is the root bending moment. It is not a time spectral stability derivative as the current description says, but it is also broken.

marcomangano:
@gang525 @ArshSaja @sseraj what is the status of this PR? I would wait for the comments from Josh and Neil to be addressed before doing another round of reviews

ArshSaja:
@gang525  can you update the branch and fix the conflicts? this happens because of the fprettify changes I think. I can fix the tapenade files once you merge the branch with main.

doc/costFunctions.yaml
```diff
@@ -0,0 +1,458 @@
...
+    This can be all wall surface or the interested surface defined in the runscript.
+
+axismoment:
+  desc: >
+    Moments about the axis given by ``momentAxis``.
+    Units: Newton - meter
+
+flowpower:
+  desc: >
+    Added power by actuator zone to the flow (f.V).
```
lamkina, line 374: What is the (f.V) here?

doc/costFunctions.yaml
```diff
@@ -0,0 +1,458 @@
...
+  desc: >
+    The separation sensor average in z direction.
+    The sensor times the distance in z direction value is computed here.
+
+cavitation:
+  desc: >
+    Cavitation sensor, not to be confused with the cavitation number.
+    It is a modified Heaviside function accounting for how much :math:`-C_p` exceeds the cavitation number over the given surface family.
+    The `computeCavitation` flag must be set to `True` because this introduces additional computations.
+    You probably will specify the surface family with `cgnsutilities`.
```
lamkina, line 313: I think we should use double (``) around "computeCavitation" and "cgnsuitlities".  Currently they show up as italicized in the docs instead of inline code.

doc/costFunctions.yaml
```diff
@@ -0,0 +1,458 @@
...
+    See :cite:t:`Liao2021a`.
+    Units: None
+
+cpmin:
+  desc: >
+    Minimum coefficient of pressure (:math:`C_p`) over the given surface family.
+    This function is computed with Kreisselmeier-Steinhauser (KS) function aggregation resulting in a conservative constraint because the `cpmin` outputted will never be more negative than the true :math:`C_{p,min}`.
```
lamkina, line 320: See comment about italicized vs. inline code for this too

doc/costFunctions.yaml
```diff
@@ -0,0 +1,458 @@
...
+  desc: >
+    Viscous lift
+
+liftmomentum:
+  desc: >
+    Momentum lift (due to changing momentum of flow in unsteady simulation)
+
+mavgvi:
+  desc: >
+    Only Anil knows the secret.
```
lamkina, line 441: Maybe change this to a `TODO` instead lol.

lamkina (changes requested):

doc/costFunctions.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+axismoment:
+  desc: >
+    Moment of the geometry in the given axis.
+
+flowpower:
+  desc: >
+    Added power by actuator zone to the flow (f.V).
+
+forcexpressure:
```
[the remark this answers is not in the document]
lamkina: This notification slipped through the cracks.  Yeah I will add a description.

lamkina (dismissed):
Thanks for fixing all the formatting stuff @gang525!  I think this is ready to go.

ArshSaja:
Thanks @gang525  for all changes! I will try to fix if anything new comes up

sseraj:
> @sseraj can you add brief descriptions to the time spectral functions and remove the "TODO"s ?

Yes, I can do that'

Issue #243, opened 2022-11-01
Support GCC 10+ 


eirikurj:
## Description 
Currently, ADflow does not compile with GCC 10 and above. The compilation fails with errors as reported in #175. This is problematic as newer compiler are becoming more available, in particular on development workstations, and docker containers based in Ubuntu 22 that are waiting to be added as [testing images](https://github.com/mdolab/docker/pull/185). 

The primary reason for this error is that the CGNS library does not provide an explicit Fortran interface at this time for some subroutines,  [see the docs](https://cgns.github.io/CGNS_docs_current/midlevel/general.html). 

## Proposed solutions
To address this I see the following approaches we can take, increasing in involvement of implementation and some pros and cons
<table>
<tr>
	<td> Suggestion
	<td> Pros
	<td> Cons
<tr>
	<td> 1. Include a comment in the relevant config files to include the <tt>-fallow-argument-mismatch</tt> flag
	<td> Minor change.
	<td> Notes have to added to the docs and/or users have to find this string in the config and add it manually. Warnings are printed while code is compiled.
<tr>
	<td> 2. Detect the compiler version and inject the <tt>-fallow-argument-mismatch</tt> flag automatically
	<td> Minimal change to build system and no user input
	<td> Warnings are still printed while code is compiled
<tr>
	<td> 3. Since we plan to support only F2008 compliant compilers, we could implement our own explicit interface shim on the CGNS library
	<td> Error removed
	<td> More source code to maintain, which I think should be done by the CGNS library itself
</table>

I dont think its worth spending the time on a solution like 3. Short term I suggest something like 2., since that will require the least effort from the user. Long term, the CGNS library will eventually offer an explicit interface, as it will require more recent Fortran standards. 

Feel free to add other suggestions or comments. We should try to address and come to a conclusion sooner than later.

PR #244, opened 2022-11-01
Allow GCC 10+ to compile

eirikurj:
## Purpose
This PR adds a script that checks the compiler version and injects the `-fallow-argument-mismatch` flag if it detects GCC version 10 and newer.  For more detail on the issue, see #243.

## Expected time until merged
No rush

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
Test compiling adflow with GCC 10+ either locally or in a container.

## Checklist

### DIFF
--- a/config/compilerCheck.sh
+++ b/config/compilerCheck.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Need one argument at least
+if [ -z $1 ]; then
+    echo "$0 needs to be called with the compiler name as an argument"; exit 1
+fi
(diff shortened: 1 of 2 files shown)


nwu63 (approved):
LGTM. The idea I had was to simply have this set in `build_real.sh` via `sed` on the `config.mk`, but this is a more automated approach.

sseraj (approved):
I checked that the compilation works with GCC 11

PR #247, opened 2022-11-28
Add center of force calculations

anilyil:
## Purpose

This PR adds the center of force computations as ADflow cost functions. I am calling this center of force and not center of pressure because we look at the sum of all forces, which is what is more relevant for RANS solvers. 

The center of force cannot be exactly determined from just the total forces and moments; however, the total moments can be computed from the center of force and total force calculations. However, I kept the moment computations as is and did not modify them. This is because the code already had a very good and granular implementation for the moment terms. It is also a good sanity check to compare the relationships between moment, total force, and center of force. 

The center of force is returned in the following functionals:
```
cofxx
cofxy
cofxz
cofyx
cofyy
cofyz
cofzx
cofzy
cofzz
```
The first coordinate determines the force direction, and the second coordinate determines the component of the center of force. For example, `cofxy` is the y-coordinate of the center of x-force. It is up to the user how these are combined and used.

I also added a cost function called `mavgvi`, which is just a derived functional based on some compressible relationships. Not useful for most people.

## Expected time until merged
Similar to #246, after #231 is merged, I need to update this PR. Currently it looks like there are tons of changes but in reality this is a standalone feature. I also need to add tests.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/reg_tests/test_center_of_force.py
+++ b/tests/reg_tests/test_center_of_force.py
@@ -80,19 +80,16 @@ class TestCenterOfForce(unittest.TestCase):
         mz = funcs[f"{self.ap.name}_mz"]
 
         # center of x-force
-        cofxx = funcs[f"{self.ap.name}_cofxx"]
         cofxy = funcs[f"{self.ap.name}_cofxy"]
         cofxz = funcs[f"{self.ap.name}_cofxz"]
 
         # center of y-force
         cofyx = funcs[f"{self.ap.name}_cofyx"]
-        cofyy = funcs[f"{self.ap.name}_cofyy"]
         cofyz = funcs[f"{self.ap.name}_cofyz"]
 
         # center of z-force
         cofzx = funcs[f"{self.ap.name}_cofzx"]
         cofzy = funcs[f"{self.ap.name}_cofzy"]
-        cofzz = funcs[f"{self.ap.name}_cofzz"]
 
         # reference point for the AP
         xref = self.ap.xRef


lamkina (commented):
~~Was there a reason you didn't add a test for the `mavgvi` cost function as well?~~
Nevermind since this shouldn't be used much.

lamkina (approved):
As far as I can tell, this looks good.

src/solver/surfaceIntegrations.F90
```diff
@@ -595,6 +675,32 @@ subroutine wallIntegrationFace(localValues, mm)
           Mv(2) = Mv(2) + my * blk
           Mv(3) = Mv(3) + mz * blk
 
+          ! the force integral for the center of pressure computation.
+          ! We need the cell centers wrt origin
+          xc = fourth*(xx(i,j,  1) + xx(i+1,j,  1) &
```
sseraj, line 680: I would changes these to `xco`, `yco`, `zco` as you call them in `flowIntegrationFace`. This helps to distinguish them from distances from the reference point.

src/solver/surfaceIntegrations.F90
```diff
@@ -449,6 +503,32 @@ subroutine wallIntegrationFace(localValues, mm)
...
+            +         xx(i,j+1,2) + xx(i+1,j+1,2))
+       zc = fourth*(xx(i,j,  3) + xx(i+1,j,  3) &
+            +         xx(i,j+1,3) + xx(i+1,j+1,3))
+
+       ! accumulate in the sums. each force component is tracked separately
+
+       ! Force-X
+       COFSumFx(1) = COFSumFx(1) + xc * fx * blk
+       COFSumFx(2) = COFSumFx(2) + yc * fx * blk
+       COFSumFx(3) = COFSumFx(3) + zc * fx * blk
```
sseraj, line 520: Do similar calculations also have to be added to zipperIntegrations.F90? The center of force test is nice, but it does not test the zipper integrations.

sseraj (commented):
Looks pretty good, but I have a few comments

tests/reg_tests/refs/actuator_tests.json
```diff
@@ -121,12 +135,13 @@
...
         "equationType": "Euler",
         "eulerWallTreatment": "linear pressure extrapolation",
+        "explicitSurfaceCallback": null,
         "firstRun": true,
         "flowType": "external",
         "forcesAsTractions": true,
         "frozenTurbulence": false,
         "globalPreconditioner": "additive Schwarz",
-        "gridFile": "input_files/actuator_test_pipe.cgns",
+        "gridFile": "tests/reg_tests/tests/reg_tests/tests/reg_tests/tests/reg_tests/tests/reg_tests/input_files/actuator_test_pipe.cgns",
```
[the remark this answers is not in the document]
sseraj: I think this was the only file with this issue.

sseraj (commented):
Looks good other than the flake8 check

sseraj (approved):

lamkina (approved):

PR #249, opened 2022-12-06
fixed sorting of evalFuncs when passed as kwargs

marcomangano:
## Purpose

The `evalFuncs` argument in `evalFunctions()` and `evalFunctionsSens()` is casted to a sorted list in several parts of `baseclasses.AeroProblem` and `pyAdflow`, except when the functions are passed as kwarg in `pyADflow.evalFuncs()` - see [this line](https://github.com/mdolab/adflow/blob/069ab3199e3caa5b33416e1fe9f6f7d0c9a8d603/adflow/pyADflow.py#L1362).
This creates issues when the `evalFuncs` kwarg is passed as a set in a multi-processor run. Python sets indeed do not preserve ordering within the iterable object, so when the function values are gathered from different procs in a for loop over the list index, the values of different functions will be mixed together.

This PR addresses this by casting the problematic list to a sorted list.
 

## Expected time until merged

Not much, I do not think a test is necessary here.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -1359,7 +1359,7 @@ class ADFLOW(AeroSolver):
         tmp = []
         for f in evalFuncs:
             tmp.append(f.lower())
-        evalFuncs = tmp
+        evalFuncs = sorted(tmp)
 
         # We need to determine how many different groupings we have,
         # since we will have to call getSolution for each *unique*


nwu63:
For reference, this is the same issue that has appeared before when we first ported things to py3. See for example mdolab/baseclasses#3 (one of my earliest PRs!)

I don't think this is the best approach though. I guess I never considered that the user would pass in a set, but if we are going to accept that, I would propose casting the input to a sorted list instead of what's in this PR. E.g. after line #1356

```
else:
    evalFuncs = sorted(list(evalFuncs))
```
which is the same approach taken in baseclasses and other repos.

marcomangano:
> For reference, this is the same issue that has appeared before when we first ported things to py3. See for example [mdolab/baseclasses#3](https://github.com/mdolab/baseclasses/pull/3) (one of my earliest PRs!)
> 
> I don't think this is the best approach though. I guess I never considered that the user would pass in a set, but if we are going to accept that, I would propose casting the input to a sorted list instead of what's in this PR. E.g. after line #1356
> 
> ```
> else:
>     evalFuncs = sorted(list(evalFuncs))
> ```
> 
> which is the same approach taken in baseclasses and other repos.

Nvm, we figured that `sorted()` already casts the set into a list. We could refactor the loop right above in a more compact way though.

sseraj (approved):
Nice work getting to the bottom of this

Issue #256, opened 2023-01-23
ADflow doesn't work with numpy 1.24


A-CGray:
## Description

Self explanatory. We are trying to update the "latest" docker images to numpy 1.24 (https://github.com/mdolab/docker/pull/208) and getting the error below in all the ADflow tests when calling `setattr`, my best guess is that something changed in f2py.

### Steps to reproduce issue

1. Install numpy v1.24
2. Run ADflow tests

### Current behavior
```text
 (mpi) /home/***/repos/adflow/tests/unit_tests/test_restart.py:BasicTests.test_import_block_splitting ... FAIL (00:00:0.03, 159 MB)
Traceback (most recent call last):
  File "/home/***/.pyenv/versions/3.10.9/lib/python3.10/site-packages/testflo/test.py", line 418, in _try_call
    func()
  File "/home/***/repos/adflow/tests/unit_tests/test_restart.py", line 33, in test_import_block_splitting
    CFDSolver = ADFLOW(options=self.options, debug=False)
  File "/home/***/.pyenv/versions/3.10.9/lib/python3.10/site-packages/adflow/pyADflow.py", line 148, in __init__
    super().__init__(
  File "/home/***/.pyenv/versions/3.10.9/lib/python3.10/site-packages/baseclasses/solvers/pyAero_solver.py", line 45, in __init__
    super().__init__(
  File "/home/***/.pyenv/versions/3.10.9/lib/python3.10/site-packages/baseclasses/solvers/BaseSolver.py", line 86, in __init__
    self.setOption(key, optionValue)
  File "/home/***/.pyenv/versions/3.10.9/lib/python3.10/site-packages/adflow/pyADflow.py", line 5207, in setOption
    setattr(module, variable, value)
ValueError: 0-th dimension must be 256 but got 0 (not defined).
```

### Expected behavior
All tests pass

sseraj:
The errors in the issue description are caused by a bug in f2py that has recently been fixed in https://github.com/numpy/numpy/pull/23194. I built numpy with the fix from source (`1.25.0.dev0+678.g51ecf8401`), recompiled ADflow, and all the tests passed. We just need to wait for the next patch release, which should include this fix. The same applies to pyHyp.

There is a separate bug that occurs in ADflow when updating from 1.21 to 1.22. This looks like the intended behavior from f2py related to `intent(out)` strings of undetermined length:

```
File "/home/mdolabuser/repos/adflow/adflow/pyADflow.py", line 5065, in setOption
[...]

PR #262, opened 2023-02-17
Fixed f2py string intent

sseraj:
## Purpose

numpy 1.22 introduced some changes to how array dimensions are determined with f2py. From what I can tell, `intent(out)` strings of undetermined length are no longer valid (or have never been valid but did not previously raise an error).

ADflow has 4 string variables with `intent(inout)` that cause problems with numpy 1.22. I posted the error message in #256. However, we never use these variables in Python after passing them to Fortran, so I changed them to `intent(in)`. This fixes the error and does not change the behavior of the code.

## Expected time until merged

1 week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I installed numpy 1.22, recompiled ADflow with these changes, and checked that all tests pass.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     packages=find_packages(include=["adflow*"]),
     package_data={"adflow": ["*.so"]},
     install_requires=[
-        "numpy>=1.16,<1.22",
+        "numpy>=1.16,<1.24",
         "mdolab-baseclasses>=1.4",
         "mpi4py>=3.0",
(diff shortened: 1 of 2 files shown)


marcomangano (approved):
This looks good, nice catch. I guess the only way to test this would be to bump numpy version on `latest` images?

marcomangano (approved):

joanibal (approved):
Nice

PR #264, opened 2023-02-21
Free stream change correction improvements

anilyil:
## Purpose

This PR adds a new way to compute a free stream correction update.

When the free stream conditions are modified for a given aeroProblem compared to the last time it was run, ADflow computes a simple correction to the states that adds the offset from the change to each cell's state. The goal here is to perturb the states enough so that the residual increases throughout the volume. If this correction is not performed, the residuals only increase near the cells that are 2 layers within the farfield BCs. As a result, the Newton-based solvers struggle to get over these transients because the residuals initially go up as the changes are propagated throughout the domain. This offset also aims to put the volume cells in a better position for convergence. The cells that have a state that is similar to the farfield conditions will be very close to their final state with the correction. This behavior is controlled with the option `infChangeCorrection`.

In this PR, I add a new method to perform this correction update. In my formulation, instead of adding the delta between the new and the old free stream conditions, I actually rotate the velocity vectors at each cell by how much the free stream velocity has rotated. This change results in a slightly better restart when the inflow angle is changed. I do not need to keep track of alpha and beta because the update is based on the directions of the old and new free stream velocities. The rotation is performed on the plane that is orthogonal to these vectors. On top of the rotation, I also add the changes to the density and energy throughout like the original method because these changes can become important with changes to altitude and Mach number.

There are now 3 options that control the correction update: 
- `"infChangeCorrection": [bool, True],` this option enables or disables this correction
- `"infChangeCorrectionTol": [float, 1e-12],` this tolerance determines when we may not want to run the correction update. If the norm of the change in the state is less than this tolerance, we dont run the update. This is useful when the state does not change, so we avoid messing stuff up by running a correction. For the offset method this is not too problematic, but the rotation needs to compute an orthogonal vector, and if the two free stream vectors are parallel, it can't and would get a nan. This tolerance prevents this from happening. The tolerance was a hard coded parameter before this PR so the PR maintains the old behavior exactly.
- `"infChangeCorrectionType": [str, ["offset", "rotate"]],` pretty self explanatory; offset does the original method, rotate rotates the velocities and offsets the density and energy.

## Expected time until merged
1-2 weeks. I want to use this feature a bit to see if it works. I also will add a test for this. 

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/doc/options.yaml
+++ b/doc/options.yaml
@@ -324,7 +324,7 @@ infChangeCorrection:
     Setting this to True will adjust the flowfield everywhere if the angle of attack changes between optimization iterations.
     The idea is that when the angle of attack is changed after a converged simulation, the residuals can stay low since this value only initially affects the residuals of the farfield boundary cells.
     As a result, the solver logic can do weird stuff because it is often based on the L2 norm of the residual.
-    If this option is set to True, we do a simple adjustment of the state vecotr with respect to the changes in the free stream conditions.
+    If this option is set to True, we do a simple adjustment of the state vector with respect to the changes in the free stream conditions.
     The method to compute the correction is determined by ``infChangeCorrectionType``.
     The correction is only run if the norm of the free stream state change is greater than ``infChangeCorrectionTol``.
     As a result, the residuals jump up and the solver logic works as designed.
@@ -332,10 +332,12 @@ infChangeCorrection:
 
 infChangeCorrectionTol:
   desc: >
-    The tolerance that determines if we will run a corraction update when the ``infChangeCorrection`` option is set to True.
+    The tolerance that determines if we will run a correction update when the ``infChangeCorrection`` option is set to True.
     We first compute the change in the free stream state vector and compute its norm.
     If this norm is greater than the tolerance set by this option, a correction update is performed.
     If the norm of the change is smaller than the tolerance, we don't perform a correction update.
+    The default value is appropriate for the default ``"offset"`` type update.
+    When the ``"rotate"`` update type is used, users might want to select a higher tolerance because the vector rotation that determines the update might be ill conditioned with small state changes.
 
 infChangeCorrectionType:
   desc: >


lamkina (commented):
A few comments to discuss before I approve this PR:
1. Should we abstract away the rotation matrix as a separate routine?  This would let us use this as a general utility in the future.
2. Does it make sense to add a test for the case where the change in the states is below the tolerance?  I think we might want to test that capability.

lamkina (dismissed):
I think this is ready to go.

ArshSaja:
It looks good to me, I have a question on the `infChangeCorrectionTol`. In which cases this tolerance can be set to a higher value? I suppose in all the cases we would set this tolerance to be a lower value, smth like `1e-12`.

doc/options.yaml
```diff
@@ -324,10 +324,25 @@ infChangeCorrection:
     Setting this to True will adjust the flowfield everywhere if the angle of attack changes between optimization iterations.
     The idea is that when the angle of attack is changed after a converged simulation, the residuals can stay low since this value only initially affects the residuals of the farfield boundary cells.
     As a result, the solver logic can do weird stuff because it is often based on the L2 norm of the residual.
-    If this option is set to True, we do a simple linear adjustment of every velocity vector in the solution with the same change in alpha.
+    If this option is set to True, we do a simple adjustment of the state vecotr with respect to the changes in the free stream conditions.
```
sseraj, line 327: vecotr -> vector

doc/options.yaml
```diff
@@ -324,10 +324,25 @@ infChangeCorrection:
...
+    The method to compute the correction is determined by ``infChangeCorrectionType``.
+    The correction is only run if the norm of the free stream state change is greater than ``infChangeCorrectionTol``.
     As a result, the residuals jump up and the solver logic works as designed.
     This supersedes the :py:data:`RKReset` approach.
 
+infChangeCorrectionTol:
+  desc: >
+    The tolerance that determines if we will run a corraction update when the ``infChangeCorrection`` option is set to True.
```
sseraj, line 335: corraction -> correction

sseraj (commented):
I only checked the docs

ArshSaja (approved):

lamkina (approved):

PR #271, opened 2023-03-03
Dissipation-based continuation for flows with shocks

sseraj:
## Purpose

This PR adds the dissipation-based continuation (DBC) method developed in [this paper](https://mdolab.engin.umich.edu/bibliography/Seraj2022c.html). The main idea is to add second order dissipation at shocks during the first several iterations to help converge supersonic flows, which are prone to stalling. The additional dissipation is reduced as the solution converges. I have been using this in my optimizations for a few months with decent success.

The formulation in this PR may eventually be replaced by @joanibal's more general DBC approach, but we decided to merge this in for now.

## Expected time until merged

2 weeks? The changes are relatively small.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added DBC to the solver combos test. I could also add a derivative test, but the changes are small enough that I don't think it is worth expanding the derivative test cases.

## Checklist


doc/options.yaml
```diff
@@ -432,6 +432,32 @@ useLinResMonitor:
...
+    This only applies to the ``central plus scalar dissipation`` :py:data:`discretization`.
+    This adds second order dissipation near shocks at the initial stages of the solution, which helps converge supersonic flows.
+    The dissipation tapers off as a sigmoid function of the relative convergence.
+    See "Improving the Performance of a Compressible RANS Solver for Low and High Mach Number Flows" (Seraj2022c) for details.
+
+dissContMagnitude:
+  desc: >
+    The magnitude of the additional dissipation for dissipation-based continuation.
+    The initial dissipation is equal to this value plus :py:data:`vis2`.
+    Values between a fifth and a half of the freestream Mach number typically work well.
```
marcomangano, line 447: The default value for this is 1, so that means that it is tuned for Mach 2?

doc/options.yaml
```diff
@@ -432,6 +432,32 @@ useLinResMonitor:
     Enables the linear residual monitor for the NK and ANK solvers.
     The output should look like the linear residual output from the adjoint solver.
 
+useDissContinuation:
+  desc: >
+    Use dissipation-based continuation.
+    This only applies to the ``central plus scalar dissipation`` :py:data:`discretization`.
+    This adds second order dissipation near shocks at the initial stages of the solution, which helps converge supersonic flows.
```
marcomangano, line 439: Should we be more explicit about the fact that this should not be used on "conventional" sub- and transonic conditions?

doc/options.yaml
```diff
@@ -432,6 +432,32 @@ useLinResMonitor:
...
+dissContMagnitude:
+  desc: >
+    The magnitude of the additional dissipation for dissipation-based continuation.
+    The initial dissipation is equal to this value plus :py:data:`vis2`.
+    Values between a fifth and a half of the freestream Mach number typically work well.
+
+dissContMidpoint:
+  desc: >
+    The number of orders of relative convergence at which the additional dissipation is half of :py:data:`dissContMagnitude`.
+    For example, a value of 1.0 means that the midpoint is at one order of magnitude of relative convergence.
```
marcomangano, line 452: rph suggestion, if it sounds better to you: "For example, `dissContMidpoint` = 1.0 means that the additional dissipation is half of the initial value by the time the solver has decreased the initial residual by one order of magnitude"

doc/options.yaml
```diff
@@ -432,6 +432,32 @@ useLinResMonitor:
...
+    The default value typically works well.
+
+dissContSharpness:
+  desc: >
+    The sharpness of the dissipation-based continuation sigmoid function.
+    Larger values will result in a steeper drop from the initial to final dissipation value.
```
marcomangano, line 458: Does this mean that the additional dissipation goes to zero sooner with higher sharpness values?

src/NKSolver/blockette.F90
```diff
@@ -3103,7 +3104,16 @@ subroutine inviscidDissFluxScalar
 
         ! Set a couple of constants for the scheme.
 
-        fis2 = rFil * vis2
+        if (useDissContinuation) then
```
marcomangano, line 3107: I would add a comment block here to reference to your paper and/or the docs? On a separate note, it would be nice to have a short description here or above about what these parameter are for future debugging. `rFil` is defined in `iteration.f90`, but it looks like the other constants are just defined above in this file without further explanation

marcomangano (commented):
As far as I understand the method, this PR looks good to me. I would just a couple of clarifications on the comments and docs. Maybe you can either copy or refer directly to Eq.(28) and Eq.(29) in your paper for clarity?

doc/options.yaml
sseraj: Thanks

doc/options.yaml
sseraj: Yes

src/NKSolver/blockette.F90
sseraj: Are you referring to `fis2`, `sfil` and others like that? They are intermediate variables, but I can try to add some descriptions.

sseraj (commented):
Thanks for the prompt review @marcomangano. I addressed some of the documentation comments. I will work on the others soon as well.

src/NKSolver/blockette.F90
sseraj: I added some comments. Can you check if this matches what you had in mind?

marcomangano (dismissed):
Thanks for adding the comments!

marcomangano (approved):

PR #274, opened 2023-03-13
Acoustic scale factor and characteristic time stepping

sseraj:
## Purpose

This PR adds the acoustic scale factor and characteristic time stepping developments from [this paper](https://mdolab.engin.umich.edu/bibliography/Seraj2022c.html).

The main changes are:
- Added `acousticScaleFactor` to the JST scheme. This scales the artificial dissipation introduced by the acoustic component of the spectral radius.
- Refactored the time step terms in the ANK solver to reduce code duplication. All time step terms are now computed in the `computeTimeStepInv` subroutine and stored in the PETSc matrix `timeStepInv`. The default ANK solver produces results that are nearly identical to before the refactor. However, I did have to retrain the error estimation test because the errors with partial convergence were slightly different.
- Added Turkel and VLR characteristic time stepping options. These modify the time step terms according to characteristic-based preconditioning.

## Expected time until merged

1 month or so. The changes to NKSolvers.F90 are substantial, so it could take some time to work through.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added the Turkel and VLR characteristic time stepping options to the solver combos test. This tests that both options work with ANK, SANK, and CSANK. I also added AGMG to the solver combos test because some of the time step changes affect this, and it was not previously tested.

I added the acoustic scale factor option to the cavitation test. This seemed like a natural fit for low Mach number dissipation scaling (even though the test is actually at Mach 0.6). This regression tests the analysis and tests the adjoint derivatives against complex step.

## Checklist


adflow/pyADflow.py
```diff
@@ -5250,6 +5250,7 @@ def _getDefaultOptions():
             ],
             "viscWallTreatment": [str, ["constant pressure extrapolation", "linear pressure extrapolation"]],
             "dissipationScalingExponent": [float, 0.67],
+            "acousticScaleFactor": [float, 1.0],
```
ArshSaja, line 5253: Do you think adding a warning will be helpful if the user sets the factor to be more than 1.0?

src/adjoint/adjointUtils.F90
```diff
@@ -179,6 +179,9 @@ subroutine setupStateResidualMatrix(matrix, useAD, usePC, useTranspose, &
 
             ! Very important to use only Second-Order dissipation for PC
             lumpedDiss = .True.
+            ! We also do not apply acoustic scaling because the dissipation stabilizes ILU
```
ArshSaja, line 182: are you saving the `acousticScaleFactor` here and avoiding any scaling here? if the `acousticScaleFactor` is set to 1 here without saving, will it propagate to the other functions where it is used?

src/NKSolver/NKSolvers.F90
```diff
@@ -77,9 +77,8 @@ module NKSolver
 
     subroutine setupNKsolver
```
ArshSaja, line 79: I just checked how the options work with the NK. I will leave @anilyil  to review the NK part.

src/NKSolver/NKSolvers.F90
```diff
@@ -2162,27 +2028,293 @@ subroutine FormJacobianANK
...
+        ! Output variables
+        real(kind=realType), dimension(nState, nState), intent(out) :: timeStepBlock
+
+        ! Local variables
+        real(kind=realType) :: blendFactor, dtInv, rho, velX, velY, velZ
+        real(kind=realType) :: speed, speedOfSound, mach, machSqr, machSqrTrunc, alpha, beta, tau, gammaMinusOne
+        real(kind=realType) :: speedXY, sinTheta, cosTheta, sinAlpha, cosAlpha
+        real(kind=realType), dimension(nState, nState) :: streamToCart, symmToCons, consToSymm, stateToCons
+
+        ! Zero the block matrices
```
ArshSaja, line 2127: By any chance, will these block matrices run out of memory?

src/adjoint/adjointUtils.F90
sseraj: `acousticScaleFactor` is saved and then restored to the user value before exiting this routine. This is handled the same way as the lumped dissipation.

adflow/pyADflow.py
sseraj: I don't think a warning would be very useful. Warnings are easy to ignore, and someone using this option should read the documentation before changing it from the default.

src/NKSolver/NKSolvers.F90
sseraj: The block matrices are 5x5 or 6x6, so they don't use much memory. Also, these block matrices are only allocated for one cell at a time.

src/adjoint/adjointUtils.F90
sseraj: I added some more comments as we discussed @anilyil

src/NKSolver/NKSolvers.F90
```diff
@@ -2378,105 +2509,9 @@ subroutine FormFunction_mf(ctx, inVec, rVec, ierr)
...
-                                    ii = ii + 1
-                                end do
-                            end do
-                        end do
-                    end do
-                end do
-            end do
-        end if
+        ! Multiply the perturbed state by the time step terms and add it to the residual
+        call MatMultAdd(timeStepInv, inVec, rVec, rVec, ierr)
```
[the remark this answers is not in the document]
sseraj: As discussed with @anilyil, I ran a case (NACA 0012 airfoil at Mach 0.5) with a constant CFL=5 and `ANKLinearSolveTol` of 1e-6 and compared the changes in this PR to the main branch. The residuals and outputs match to 6 digits for each nonlinear iteration. I think this proves that the time steps terms are the same as before. I have attached the output here for reference.

[main.log](https://github.com/mdolab/adflow/files/11467515/main.log)
[cts.log](https://github.com/mdolab/adflow/files/11467516/cts.log)

tests/reg_tests/refs/error_euler_wing_partial.json
```diff
@@ -1,12 +1,12 @@
 {
     "errors": {
```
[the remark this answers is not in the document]
nwu63: Yes this is expected, since this PR changes the path of convergence and the test is checking some semi-converged state. This test is probably poorly written, it would better to do the following:
- partial convergence
  - compute error, then fully converge the solution and check that the error estimation is close to the real error
- full convergence
  - check errors are almost zero (we already have this)

That way the test is independent of the semi-converged state as long as we are in the linear region of convergence.

tests/reg_tests/test_solver_combos.py
```diff
@@ -83,6 +83,33 @@
             "dissContMagnitude": 0.1,
         },
     },
+    # RANS test with AGMG preconditioner
```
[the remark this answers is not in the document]
sseraj: The solver was only broken for AGMGLevels > 1, which is why this test was working. This test will not work with AGMGLevels > 1 though because the mesh is too coarse. AGMG should work the same now compared to the main branch as I mentioned in the other comment.

src/NKSolver/NKSolvers.F90
```diff
@@ -4147,6 +4086,9 @@ subroutine ANKStep(firstCall)
             lambda = zero
         end if
 
+        ! Update the inverse time-step matrix
```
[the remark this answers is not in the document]
sseraj: Thanks for the suggestion. I moved the time step computation into the if check for the PC and added some comments. I also checked that the steps the solver takes are the same as before this change.

ArshSaja (commented):
Looks good to me, Nice work @sseraj !

ArshSaja (approved):

PR #277, opened 2023-03-27
Minor  changes for ANK solver behavior

anilyil:
## Purpose

This PR contains the ANK-related changes from #246. The CL solver in that PR will need more work, but I want to clean up some of my work branches sooner than that. 

The changes are:
1. ANK CFL Reset: I added a python option `"ANKCFLReset"` that defaults to `True`. This was the original behavior of the solver, where at each subsequent ADflow call, the ANK CFL would be reset to `"ANKCFL0"` and then the moving lower bound was applied in the solver itself based on the L2 convergence. Now, the users can set `"ANKCFLReset"` to `False`, so that the last ANK CFL is preserved exactly as the solver goes into the next iteration. This improves performance near the final stages of optimizations, or with CL-solvers where the problem does not change all that much. However, restarting the new case at high CFL numbers can make the solver unstable, so I am keeping the original behavior as default. The ANK CFL for each case is saved in the aeroProblem's ADflow data.

2. ANK Linear Solver Buffer: This was a hard-coded value in fortran, which is now exposed to Python. This determines how much we over-solve the linear system when we are near the L2 convergence target. Not really relevant for most people, only really comes in handy when you are trying to do performance tests and you would prefer the linear solver to quit as soon as you think you are near the L2 target. This is a linear approximation, so achieving the L2 target is not guaranteed. That's why we multiply the L2 target by the linear solver buffer `"ANKLinearSolveBuffer"`, which defaults to 1e-2. Increasing this value to something like 0.5 would be a bit more accurate at the risk of running a few extra iterations due to the linear approximation not being an exact calculation of the nonlinear residual.

3. Added a new PC update tolerance: The `"ANKPCUpdateTolAfterCutoff"` option can be used to control the PC update based on L2 convergence after the `"ANKPCUpdateCutoff"` is reached. Again, mostly a performance related option that most people don't need to worry about.

4. Added method to get number of cells: The `getNCells` method in pyADflow returns the total number of  cells in the grid as computed in the fortran layer.

## Expected time until merged

Would be great if this is done within a week.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


doc/options.yaml
```diff
@@ -1049,6 +1067,11 @@ ANKPCUpdateCutoff:
     This option only changes the PC update behavior with the coupled ANK variants.
     Decoupled ANK variants don't get affected by this.
 
+ANKPCUpdateTolAfterCutoff:
+  desc: >
+    The ANK PC update tolerance used after we reach the ``ANKPCUpdateCutoff`` convergence.
```
sseraj, line 1072: Can you also change the description for `ANKPCUpdateCutoff`? I think the parts about disabling the update should be removed and it should just say that a different tolerance is used.

sseraj (commented):
Looks pretty good. Just some minor comments

sseraj (approved):

adflow/pyADflow.py
```diff
@@ -3128,6 +3149,10 @@ def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True):
             if releaseAdjointMemory:
                 self.releaseAdjointMemory()
 
+        if not self.getOption("ANKCFLReset"):
+            # also set the ANK CFL to the last value if we are restarting
+            self.adflow.anksolver.ank_cfl = aeroProblem.adflowData.ank_cfl
```
marcomangano, line 3154: I assume this just takes the initial default value in the fortran layer on the first iteration when `ANKCFLReset: False`?
(It looks like it from the added id statement in the fortran layer for `ANK_CFL = ANK_CFL0` but I want to double check)

doc/options.yaml
```diff
@@ -951,6 +951,18 @@ ANKLinearSolveTol:
...
+    The second scenario is what this option is useful for; this option is used to further offset the absolute linear solver tolerance in the ANK solver.
+    This option only really makes a difference very close to the final L2 convergence target.
+    For example, setting a value of 0.01 will force the linear solver to converge the linear system more tightly, in the cases where the absolute linear convergence is achieved before the relative target.
+    Reducing the value here to 1e-4 or 1e-6 may result in "extra" nonlinear convergence in the final iteration, at the cost of more linear solver effort.
+    Increasing the value here to 1e-1 will reduce this buffer and the solver is more likely to end closer to the final L2 convergence target, but a too large of a value might cause issues with the final iterations where several smaller additional Newton steps are required.
```
marcomangano, line 964: Maybe worth mentioning the "performance improvement" motivation here as well?

adflow/pyADflow.py
marcomangano: Yeah my phrasing was odd but that is what I meant, sounds good

marcomangano (approved):

sseraj (approved):

PR #280, opened 2023-04-23
Fix type bug pyf file

eirikurj:
## Purpose
This PR addresses a very subtle bug causing unspecified behavior when running in complex mode. This bug was uncovered while debugging the new Intel OneAPI compilers. 

Several parameters are defined as `real(kind=alwaysrealtype)` in fortran, but as `real(kind=realtype)` in the python interface. When complexifying, these are changed as specified in the pyf file (incorrectly) to `complex(kind=8)`. This means that when these particular variables are set from python, 8 bytes are set in fortran, i.e., the first 4 bytes (the actual variable), and the adjacent 4 bytes in memory, which in principle can be anything. In practice, it turned out to be overwriting some other options. 

In the case of the new Intel compilers, `vis4` was overwritten in fortran, after the option had been set correctly earlier, causing a strange convergence behavior. On the GCC images, the booleans `useALE` and `updateWallDistanceUnsteady` are changed to false. The tests did not catch this, as they are both time-accurate options and current regression tests do not include complex tests. 

To prevent this happening again, a unit test is added to check that the parameters that are set from python are set to the expected value in fortran after the initialization of the solver. 

## Expected time until merged
Few days.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Run the included unit test without and with the fix.

## Checklist

### DIFF
--- a/tests/unit_tests/test_basics.py
+++ b/tests/unit_tests/test_basics.py
@@ -1,6 +1,8 @@
 import unittest
 import os
+import numpy as np
 from adflow import ADFLOW
+from adflow import ADFLOW_C
 
 
 baseDir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_basics.py
```diff
@@ -14,6 +16,70 @@ def test_import(self):
...
+                # we just overwrite the python value. An example is "solutionprecision" which is
+                # a str value in python that corresponds to an integer in fortran, which sets a module variable precisionsol
+                module = CFDSolver.moduleMap[CFDSolver.optionMap[optionName]["location"][0]]
+                variable = CFDSolver.optionMap[optionName]["location"][1]
+                value = CFDSolver.optionMap[optionName][value.lower()]
+            else:
+                module = CFDSolver.moduleMap[CFDSolver.optionMap[optionName][0]]
+                variable = CFDSolver.optionMap[optionName][1]
+
+            # Fetch the variable value form the module
```
sseraj, line 44: form -> from

tests/unit_tests/test_basics.py
```diff
@@ -14,6 +16,70 @@ def test_import(self):
...
+                fortranValue = str(np.array(fortranValue, dtype=str)).strip()
+
+            elif optionName == "mgstartlevel" and (value == -1 or value == 0):
+                # mgstartlevel is set in Fortran but the default value (-1) defaults to the coarsest level
+                # value of -1 or 0 should default to coarsest level, which is extracted based on the number of MG levels strategy
+
+                # Just overwrite value with what is set in fortran
+                value = CFDSolver.adflow.inputiteration.nmglevels
+
+            # Note/todo: When running in complex mode the real part is compared to the default value. Options should in most cases always be real, but are complex in fortran. This should be updated at some point.
```
sseraj, line 62: Can you clarify what this means? Are you saying that we have options that should be alwaysRealType but are realType instead?

sseraj (commented):
Nice work! Just two minor comments

tests/unit_tests/test_basics.py
eirikurj: fixed

tests/unit_tests/test_basics.py
eirikurj: I updated the comment as it was not entirely accurate. It suggested that most options are of incorrect type, which is not the case. There are some options that could be changed. For example, `loadImbalance` does probably not need to be `realType`. However, there is no harm in keeping them `realType`, as long as its handled properly in the code.

tests/unit_tests/test_basics.py
```diff
@@ -14,6 +16,70 @@ def test_import(self):
...
+            # Now check if the set parameter is correct in fortran
+            if type(value) is bool:
+                fortranValue = bool(fortranValue)
+
+            elif type(value) is str:
+                # To not mess string arrays too much the following converts string byte array to regular array and strip whitespace
+                fortranValue = str(np.array(fortranValue, dtype=str)).strip()
+
+            elif optionName == "mgstartlevel" and (value == -1 or value == 0):
+                # mgstartlevel is set in fortran but the default value un python (-1) defaults to the coarsest MG level in fortran (based on the MG levels strategy)
```
sseraj, line 55: Typo un -> in

eirikurj: fixed

sseraj (approved):
Thanks!

PR #296, opened 2023-07-08
Slice mods

anilyil:
## Purpose

This PR adds the first interface that lets users add "active" slices and query the current slice data in ADflow through memory. 

`addActiveSlice` method in pyADflow is used to define the active slices; one at a time for now.

`getActiveSliceData` method in pyADflow returns all of the active slice data in a dictionary. 

Ultimately, we will want to process this data in the dict by directly passing it to the airfoil classes, and we can just bookkeep the airfoil class instances in pyADflow.

I am attaching a script to run the example code. This is a simplified script from our airfoil tutorial, with the relevant naca0012 mesh.

[active_slice_test_files.zip](https://github.com/mdolab/adflow/files/11991325/active_slice_test_files.zip)

### DIFF
--- a/adflow/airfoilClass.py
+++ b/adflow/airfoilClass.py
@@ -642,6 +642,31 @@ class Airfoil:
         fig.update_scenes(aspectmode="data")
         fig.show()
 
+    def addFunction(self, funcName, callBack):
+
+    def getNodeFiltering(self, begFrac, endFrac, surface="upper"):
+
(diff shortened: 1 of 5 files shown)


lamkina:
@gawng I think this effort is currently stalled and I don't think there's a timeline on anyone continuing this effort at the moment. I think it's best to leave this as a draft for now.

PR #299, opened 2023-07-20
Custom pointsets for dvgeo

anilyil:
## Purpose

This PR adds the capability to specify custom pointsets based on family names that are tracked separately in DVGeo. Each pointset can receive its own custom kwargs. See the relevant PR in baseclasses for how the custom families are specified: https://github.com/mdolab/baseclasses/pull/89

This is a somewhat complete implementation but there are several assumptions. First, the custom families should not overlap. The implication here comes up with derivative computations; if families overlap, the surface seeds of the overlapped nodes will be double counted.

Second, we assume the custom families span the selected design family, although this is not a hard requirement. We may want to add a check if the provided families actually do span the entire design family.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


adflow/pyADflow.py
```diff
@@ -4523,6 +4595,7 @@ def computeJacobianVectorProductFwd(
         # For the geometric xDvDot perturbation we accumulate into the
         # already existing (and possibly nonzero) xsdot and xvdot
         if xDvDot is not None or xSDot is not None:
+            # TODO fix
```
A-CGray, line 4598: Still needs fixing?

A-CGray (changes requested):
Hate to be that guy, but is there a test we can add for this?

Also, there are a lot of if else blocks added here that bascially do the same thing, either once for the design family group or once for each custom point set family. Couldn't we avoid all this additional logic if we treated the design family group as a custom point set family?

anilyil:
Converting this to a draft because we have several things we need to do that changes this part of the code. I will re-visit this once we can complete the feature.

PR #300, opened 2023-07-20
fixing petsc output vector error in the MPhys wrapper

ArshSaja:
## Purpose

In rev, when the adjoint function is called, there is some issue with f2py handling some outputs in the function:. For example,  `d_residuals["adflow_states"]` is not passing the correct output from the Fortran layer to Python in this call: `solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], d_residuals["adflow_states"], True)`.  Thus, Some derivatives become wrong:

```
Full Model: 'cruise0.coupling.prop.Fn' wrt 'geo_dvs.Nacelle:XSecCurve_1:Circle_Diameter'
    Analytic Magnitude: 2.563607e+02
          Fd Magnitude: 4.547066e+02 (fd:None)
    Absolute Error (Jan - Jfd) : 1.983459e+02 *

    Relative Error (Jan - Jfd) / Jfd : 4.362064e-01 *
    MPI Rank 8

    Raw Analytic Derivative (Jfor)
[[-256.36070479]]

    Raw FD Derivative (Jfd)
[[-454.70664024]]
```

This PR address this issue with a minor fix. It fixes the derivative issues.
```
  Full Model: 'cruise0.coupling.prop.Fn' wrt 'geo_dvs.Nacelle:XSecCurve_1:Circle_Diameter'
    Analytic Magnitude: 4.556590e+02
          Fd Magnitude: 4.547229e+02 (fd:None)
    Absolute Error (Jan - Jfd) : 9.361134e-01 *

    Relative Error (Jan - Jfd) / Jfd : 2.058646e-03 *
    MPI Rank 0

    Raw Analytic Derivative (Jfor)
[[-455.6590081]]

    Raw FD Derivative (Jfd)
[[-454.72289467]]
```

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/adflow/mphys/mphys_adflow.py
+++ b/adflow/mphys/mphys_adflow.py
@@ -470,8 +470,10 @@ class ADflowSolver(ImplicitComponent):
         if mode == "fwd":
             d_outputs["adflow_states"] = solver.solveDirectForRHS(d_residuals["adflow_states"])
         elif mode == "rev":
+            phi = np.zeros_like(d_residuals["adflow_states"])
             # d_residuals['adflow_states'] = solver.solveAdjointForRHS(d_outputs['adflow_states'])
-            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], d_residuals["adflow_states"], True)
+            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], phi, True)
+            d_residuals["adflow_states"] = phi
 
         return True, 0, 0
 


adflow/mphys/mphys_adflow.py
```diff
@@ -470,8 +470,10 @@ def solve_linear(self, d_outputs, d_residuals, mode):
         if mode == "fwd":
             d_outputs["adflow_states"] = solver.solveDirectForRHS(d_residuals["adflow_states"])
         elif mode == "rev":
+            phi = np.zeros_like(d_residuals["adflow_states"])
             # d_residuals['adflow_states'] = solver.solveAdjointForRHS(d_outputs['adflow_states'])
```
A-CGray, line 474: Remove this line if not needed

A-CGray (changes requested):

adflow/mphys/mphys_adflow.py
```diff
@@ -470,8 +470,10 @@ def solve_linear(self, d_outputs, d_residuals, mode):
...
-            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], d_residuals["adflow_states"], True)
+            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], phi, True)
```
A-CGray, line 475: Remove the commented out line if it's not needed

ArshSaja: Fixed it.

adflow/mphys/mphys_adflow.py
```diff
@@ -608,8 +608,9 @@ def solve_linear(self, d_outputs, d_residuals, mode):
         if mode == "fwd":
             d_outputs["adflow_states"] = solver.solveDirectForRHS(d_residuals["adflow_states"])
         elif mode == "rev":
-            # d_residuals['adflow_states'] = solver.solveAdjointForRHS(d_outputs['adflow_states'])
-            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], d_residuals["adflow_states"], True)
+            phi = np.zeros_like(d_residuals["adflow_states"])
```
sseraj, line 611: Is this turning off the adjoint restart? In pyADflow, we have an option for turning off the restart:
https://github.com/mdolab/adflow/blob/4dda7b34954b066f6b3dabd65abad2aeeed9259e/adflow/pyADflow.py#L4060-L4065

adflow/mphys/mphys_adflow.py
```diff
@@ -608,8 +608,9 @@ def solve_linear(self, d_outputs, d_residuals, mode):
         if mode == "fwd":
             d_outputs["adflow_states"] = solver.solveDirectForRHS(d_residuals["adflow_states"])
         elif mode == "rev":
-            # d_residuals['adflow_states'] = solver.solveAdjointForRHS(d_outputs['adflow_states'])
-            solver.adflow.adjointapi.solveadjoint(d_outputs["adflow_states"], d_residuals["adflow_states"], True)
+            phi = d_residuals["adflow_states"]
```
A-CGray, line 611: Don't you need to make a copy here?

ArshSaja: Good Catch!

A-CGray (approved):

ArshSaja:
> Do you really need both of the copies? I would imagine the openmdao vector is already doing one copy when setting the values? Did you test that the second copy is strictly necessary?
> 
> Also, any idea why just passing the pointer did not work? Maybe we can get rid of all of the copies?

Actually, the second copy is not necessary. I just did a copy for consistency. The first one is necessary though. Just setting the pointer didn't work. I am okay with removing it too

ArshSaja:
> Remove the second copy if its not strictly necessary. Try to understand why the first copy is needed. We may be able to get away with not copying it. Might need to slightly modify the f2py wrapper or add a new wrapper function that takes in the openmdao vector.

I fixed it. I think this is the only place we are using `copy`, for now, we can just do the copy may be. later if it's necessary, we can just add a helper function to set them properly.

ArshSaja:
> To avoid further delaying this fix, I say this: test it again with the current state. Make sure all relevant cases work. Then, we can merge this PR and create an issue to investigate the removal of the first copy later.

It tested and works fine with the aeroprop cases!

PR #302, opened 2023-08-01
Updating base makefile to clean complex build properly

eirikurj:
## Purpose
This PR removes the local complexify script and module and uses the shared [complexify library](https://github.com/mdolab/complexify). Docs are also updated to reflect these changes. 

## Expected time until merged
1 week

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


sseraj:
`test_solver_combos.py:TestCmplxSolverCombos_7_rans_smoother_ank_sank_csank_nk_turbksp_agmg.cmplx_test_convergence` has failed on the buildbot before as well. I think the solution is to increase nCycles a bit.

eirikurj:
Right, this test failed recently on the [tacc image](https://dev.azure.com/mdolab/Public/_build/results?buildId=6386&view=logs&j=62539d52-d31b-543b-7a48-2e344afa81aa&t=81d36c24-dfe8-5baf-7e33-8371cf41260c&l=1023), but that was the real test. Do you think there is some randomness or uninitialized variables, causing this to fail at random? If thats the case, `ncycles` would just mask the problem. I restarted the test, just to see if it will pass on the second try.

sseraj:
> Do you think there is some randomness or uninitialized variables, causing this to fail at random?

This is possible. It seems like the randomness only occurs with Intel MPI. I don't think increasing `nCycles` will mask any serious problems though. The solution path is nondeterministic with Intel sometimes.

sseraj (approved):

PR #307, opened 2023-08-19
Turning off extraneous aeroProblem print outs from MPhys wrapper

lamkina:
## Purpose
This PR adds arguments to `setAeroproblem` and `_setAeroProblemData` in `pyADflow` to turn off printing of angle of attach and BC warnings.  ~~To turn off boundary condition warning messages, a new logical argument was added to `setBCDataFineGrid` subroutine in the `BCData` module.  The usage of the `setBCDataFineGrid` in other subroutines was updated to maintain current functionality.~~

Now we turn off the boundary conition printing with fortran module variables.

~~The docstrings for `setAeroproblem` and `_setAeroProblemData` were also updated.~~
The docstrings were set back to their previous state now that we use a module variable.

An example of the printouts that happen everytime `setAeroproblem` is called is shown below.  This happens during the coupled adjoint and primal solve.  The angle of attack and boundary condition warnings are printed several times because of the extra calls required in MPhys, even though sometimes we are not actually switching the aeroproblem.

```
============
TOC.coupling
============
-> Alpha... 0.000000 
-> Alpha... 0.000000 
LN: LNBGS 0 ; 2006628.67 1
-> Alpha... 0.000000 
Solving linear in mphys_adflow
Solving ADjoint Transpose with PETSc...
   0 KSP Residual norm 9.1932257536E+03
  10 KSP Residual norm 9.1898236124E+03
  20 KSP Residual norm 9.1880786747E+03
  30 KSP Residual norm 9.1559423727E+03
  40 KSP Residual norm 9.0338778426E+03
  50 KSP Residual norm 8.3380256715E+03
  60 KSP Residual norm 6.1930959481E+03
  70 KSP Residual norm 3.5505594710E+03
  80 KSP Residual norm 1.4910352070E+03
  90 KSP Residual norm 6.3038686621E+02
 100 KSP Residual norm 2.1666819396E+02
 110 KSP Residual norm 7.8767925855E+01
 120 KSP Residual norm 2.2629509372E+01
 130 KSP Residual norm 6.3923943385E+00
 140 KSP Residual norm 1.6993385453E+00
 150 KSP Residual norm 4.0424036578E-01
 160 KSP Residual norm 1.3088391770E-01
Solving ADjoint Transpose with PETSc time (s) =    37.11
 Norm of error = 8.1977E-02    Iterations =  164
 ------------------------------------------------
 PETSc solver converged after   164 iterations.
 ------------------------------------------------
#
#*==================== !!! Warning !!! ======================
# Subsonic inflow regions present for which the turbulence
# quantities are not or insufficiently prescribed.
# Using free stream values instead.
#*===========================================================
#
#
#*==================== !!! Warning !!! ======================
# Subsonic inflow regions present for which the turbulence
# quantities are not or insufficiently prescribed.
# Using free stream values instead.
#*===========================================================
#
#
#*==================== !!! Warning !!! ======================
# Subsonic inflow regions present for which the turbulence
# quantities are not or insufficiently prescribed.
# Using free stream values instead.
#*===========================================================
#
#
#*==================== !!! Warning !!! ======================
# Subsonic inflow regions present for which the turbulence
# quantities are not or insufficiently prescribed.
# Using free stream values instead.
#*===========================================================
#
-> Alpha... 0.000000 
-> Alpha... 0.000000 
-> Alpha... 0.000000 
LN: LNBGS 1 ; 8.99476042 4.48252363e-06
-> Alpha... 0.000000 
Solving linear in mphys_adflow
Solving ADjoint Transpose with PETSc...
   0 KSP Residual norm 9.4033593773E+01
  10 KSP Residual norm 9.3851727433E+01
  20 KSP Residual norm 9.3812444502E+01
  30 KSP Residual norm 9.3662893414E+01
  40 KSP Residual norm 9.2378872212E+01
  50 KSP Residual norm 8.8436313339E+01
  60 KSP Residual norm 5.1510695403E+01
  70 KSP Residual norm 1.7465079446E+01
  80 KSP Residual norm 7.2535328261E+00
  90 KSP Residual norm 3.7911121457E+00
 100 KSP Residual norm 1.1015529443E+00
 110 KSP Residual norm 2.6571294463E-01
 120 KSP Residual norm 7.1421136189E-02
 130 KSP Residual norm 1.7754931595E-02
 140 KSP Residual norm 5.2134242805E-03
 150 KSP Residual norm 2.0794183305E-03
Solving ADjoint Transpose with PETSc time (s) =    35.69
 Norm of error = 8.7121E-04    Iterations =  158
 ------------------------------------------------
 PETSc solver converged after   158 iterations.
```

## Expected time until merged
1 week

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Only visual testing.  I ran runscripts with and without the change to ensure the printing flags were working correctly.

## Checklist


A-CGray (approved):

marcomangano (commented):
LGTM, could you just update the description by adding a printout example for reference?

lamkina:
I just updated the description.  I realized I turned off printing in `solve_linear` in the `ADflowSolver` component.  Once I switch that back on this should be ready for another review.

adflow/pyADflow.py
```diff
@@ -3282,8 +3282,18 @@ def setSurfaceCoordinates(self, coordinates, groupName=None):
         )
         self.mesh.setSurfaceCoordinates(meshSurfCoords)
 
-    def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True):
-        """Set the supplied aeroProblem to be used in ADflow"""
+    def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True, printInfo=True):
```
eirikurj, line 3285: Is there a need to set `printInfo` on every `setAeroProblem`? I would prefer this to be controlled through the options, similar to other print variables (with a more descriptive name though), and not having to pass this through a stack of functions.

marcomangano (dismissed):

adflow/pyADflow.py
```diff
@@ -3658,7 +3677,7 @@ def _setAeroProblemData(self, aeroProblem, firstCall=False):
             self.adflow.bcdata.setbcdata(nameArray, dataArray, groupArray, 1)
 
         if not firstCall:
-            self.adflow.initializeflow.updatebcdataalllevels()
+            self.adflow.initializeflow.updatebcdataalllevels(printInfo)
```
eirikurj, line 3680: Also here, instead of passing boolean values here (and elsewhere in Fortran), turning off BC warning should be done and controlled by an option, which sets a module variable in the Fortran layer. 
Then its just an `use` in the subroutines that need access to this flag.
There already exists a [similar variable](https://github.com/mdolab/adflow/blob/e7ada3de62b6500b574545b4f1c7f04c01676bf9/src/modules/inputParam.F90#L286C16-L286C29) `printWarnings`  which is similar in function, but for negative volumes. I think a `printBCWarnings` variable (or similar) would make more sense here.

eirikurj (changes requested):

A-CGray:
Bump on this

adflow/pyADflow.py
eirikurj: Sorry for the delay on this.  
I can see that turning stdout printing on/off for various things might be useful in some cases. Controlling the `alpha` using an option is perhaps a bit excessive. I do think though the BCData warnings should probably not be turned off in general, unless you really have a good reason to, as they most often indicate that something is not set properly or potential issues. Overall, there is probably a small demand for creating an option for the end user to disable the BC data warnings. 

However, setting user available option issue aside, I think the static `True/.true.` arguments to functions in various places (both python and fortran) are not great and make the code harder to read and less maintainable. The same functionality can be achieved by making (and exposing in python) a boolean module variable (see [the below comment](https://github.com/mdolab/adflow/pull/307/files#r1304459506) for the variable discussion) that controls the BCData warnings. This variable is then simply set (and reset) when calling the `setAeroProblem` in the mphys wrapper, or alternatively, in the `_setAeroProblemData` if you prefer using the `printInfo` keyword. I think this will make the code more maintainable, as there is no need to carry/set flags in multiple places. 
Let me know if this is unclear, and we can discuss, and perhaps others have further suggestions.

lamkina:
@eirikurj can you review my response about passing an argument vs. an option and let me know what you think?

eirikurj:
@lamkina I just posted a response today, it should already be there https://github.com/mdolab/adflow/pull/307#discussion_r1320056229.

lamkina:
> @lamkina I just posted a response today, it should already be there [#307 (comment)](https://github.com/mdolab/adflow/pull/307#discussion_r1320056229).

Sorry I missed the notification.  Thanks!

lamkina:
@eirikurj @A-CGray @marcomangano  I updated this PR based on the feedback to switch to option and module variables to control the printing.  I removed all the arguments and replaced them with a module variable in the fortran layer that we only call in MPhys called `printbcwarnings`.  I did not expose this as a global ADflow option because it's an MPhys only use case that normal MACH users shouldn't touch.

I also added some new logic to determine when to print the BC warnings and alpha updates in the mphys layer in `setAeroProblem`.  In a nutshell, it checks if the AP changed.  If the AP changes we always print the information.  If the AP didn't change, it then checks if we are doing a new solver iteration.  If the iteration is different for the same AP, then it prints the information (i.e single point opt with the same aeroproblem).

adflow/mphys/mphys_adflow.py
```diff
@@ -178,7 +196,10 @@ def setAeroProblem(solver, ap, ap_vars, inputs=None, outputs=None, print_dict=Tr
 
         updatesMade = set_vol_coords(solver, inputs)
 
-    if outputs is not None:
+    # We need to make sure this isn't the first optimization iteration
+    # because OM will overwrite the ADflow state vector with all ones
+    # before the first iteration.
+    if outputs is not None and ap.adflowData.callCounter != -1:
         updatesMade = set_states(solver, outputs)
```
A-CGray, line 203: I guess this is fine for now, but in future a better approach would be to find a way to actually set the initial value of the OpenMDAO outputs to the freestream initial conditions used by ADflow (can we do this during setup or configuration?). This approach of ignoring OpenMDAO inputs and doing stuff behind its back always gives me a bad feeling

lamkina: I agree I think we should be able to do this without going behind OM's back as well.  I will see if I can figure something out.

A-CGray: Think it might not be possible to do it by directly specifying the values when calling `add_output` because that is called in the `setup` where we don't even know the state vector size yet right? Let alone the freestream conditions.
https://github.com/mdolab/adflow/blob/820f182d6ab2722ba92d0c41ea8c459d3a4095e6/adflow/mphys/mphys_adflow.py#L400

Could we do it using the `guess_nonlinear` method?

lamkina: We could potentially do it during guess nonlinear, however I think that is called everytime solve_nonlinear is called, so we would need logic to protect warm starts within coupled NLBGS sovler loops.

lamkina: I just removed the hacky logic we were using in `setAeroProblem` in the mphys layer to prevent OM from overwriting the initial states.  I will open a separate PR that addresses that issue without going behind OM's back.

A-CGray:
Can you fix this conflict @lamkina , then I will review

lamkina:
> Can you fix this conflict @lamkina , then I will review

Oh yeah, just missed it when I was updating the state stuff.  I'll take care of this now

lamkina:
Should be fixed now.

lamkina:
@A-CGray this should be ready for review now.

adflow/mphys/mphys_adflow.py
```diff
@@ -955,8 +960,14 @@ def nom_write_solution(self, **kwargs):
         ap = self.ap
 
         # re-set the AP so that we are sure state is updated
+        solver.setOption("printIterations", False)
+        solver.adflow.inputiteration.printbcwarnings = False  # Turn of extra printouts
```
A-CGray, line 964: off* :P

adflow/pyADflow.py
```diff
@@ -3660,7 +3681,7 @@ def _setAeroProblemData(self, aeroProblem, firstCall=False):
             self.adflow.bcdata.setbcdata(nameArray, dataArray, groupArray, 1)
 
         if not firstCall:
-            self.adflow.initializeflow.updatebcdataalllevels()
+            self.adflow.initializeflow.updatebcdataalllevels
```
A-CGray, line 3684: Why doesn't this need brackets? Is this some weird fortran thing I don't know about?

A-CGray (changes requested):

adflow/pyADflow.py
lamkina: It should have brackets.  I must've delted them accidentally and that's probably why the tests are all failing

A-CGray: I think there was another instance of this in the PR too

A-CGray (dismissed):

adflow/mphys/mphys_adflow.py
```diff
@@ -232,7 +232,7 @@ def mphys_get_triangulated_surface(self, groupName=None):
         return self._getTriangulatedMeshSurface(groupName=groupName)
 
     def _getTriangulatedMeshSurface(self, groupName=None, **kwargs):
-        """
+        """©
```
A-CGray, line 239: Is this function copyrighted?

lamkina: Whoops

A-CGray (dismissed):
I ran this myself and it does what it's supposed to do

A-CGray:
@marcomangano @ArshSaja this is good to go

A-CGray:
@marcomangano @ArshSaja bump

lamkina:
@ArshSaja @marcomangano bump again

sseraj:
@eirikurj needs to approve because he requested changes

eirikurj (changes requested):
Sorry for the delay. Added few minor comments.

adflow/pyADflow.py
```diff
@@ -3285,7 +3291,15 @@ def setSurfaceCoordinates(self, coordinates, groupName=None):
         self.mesh.setSurfaceCoordinates(meshSurfCoords)
 
     def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True):
-        """Set the supplied aeroProblem to be used in ADflow"""
+        """Set the supplied aeroProblem to be used in ADflow.
+
+        Parameters
+        ----------
+        aeroProblem : AeroProblem
```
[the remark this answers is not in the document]
lamkina: How do you do get this to work in a docstring?  I figured out how to do it for rst files, but when I use that format in the docstring it just prints the static text.

adflow/pyADflow.py
```diff
@@ -93,6 +93,12 @@ def __init__(self, comm=None, options=None, debug=False, dtype="d"):
             curDir = os.path.basename(os.path.dirname(os.path.realpath(__file__)))
             self.adflow = MExt.MExt("libadflow", curDir, debug=debug)._module
 
+        # This is a flag not exposed to the user as an option that controls
+        # printing the boundary condition warnings.  This should only
+        # be used in the MPhys layer when we make multiple calls to
+        # updateBCAllLevels.
+        self.adflow.inputiteration.printbcwarnings = True
```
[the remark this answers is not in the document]
lamkina: Good call.  This should be taken care of now.

adflow/pyADflow.py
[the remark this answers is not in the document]
eirikurj: Yeah this is a little cryptic, but try,
```
aeroProblem : :class:`baseclasses:baseclasses.problems.pyAero_problem.AeroProblem`
```
which is an explicit link. For a shorthand render of just the classname `AeroProblem` then add a tilde in front, e.g.,
```
aeroProblem : :class:`~baseclasses:baseclasses.problems.pyAero_problem.AeroProblem`
```
See if this works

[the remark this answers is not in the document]
lamkina: I updated all the docstrings with the aeroproblem to use this linking.

A-CGray (dismissed):
If @eirikurj 's happy with this PR then I am too

eirikurj (approved):
LGTM, thanks for the effort on this. BTW, instead of requesting changes, I just adjusted the case on the Fortran variable to be consistent.

A-CGray (approved):

A-CGray:
I approve but don't have permission to merge

PR #316, opened 2023-09-09
Add center of lift functional

anilyil:
## Purpose

This PR adds the coordinates of center of lift as a functional in ADflow. This could be computed using the forces and  centers of forces; however, getting a differentiated version in the runscripts would require unnecessary headaches due to the sensitivity of this to the lift direction. 

Now, the functional is computed based on a weighed sum of the contributions of individual forces' contribution to lift. Same thing can be done for drag, but not sure how useful that would be. 

I added this functional to the regular tests. I will be re-training the tests, but we may need someone with an x86 CPU to re-train them again. 

## Expected time until merged

No rush

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


lamkina (approved):
This looks ready to me.

sseraj (commented):
This looks fine, but do these changes also need to be added to zipperIntegrations.F90?

sseraj (approved):
@anilyil You are right that this does not need to be added to zipperIntegrations.F90 because the changes are in `getCostFunctions` and not in the actual integration routines.

PR #323, opened 2023-09-29
Fixed rotation rate for local blocks

shubham135sd:
## Purpose
Fixes `setRotationRate(rotCenter, rotRate, cgnsBlocks=None)`. Currently when cgnsBlocks are passed, adflow adds rotation rate to 'n' cgnsBlocks starting from the first block number instead of applying the rotation rate to the specified block ids. The changes in this PR fix this bug.

## Expected time until merged
One week

## Type of change

## Testing
Proper test case to be created.

## Checklist
- [ ] I have tested locally that my fix is effective or that my feature works

### DIFF
--- a/adflow/pyADflow.py
+++ b/adflow/pyADflow.py
@@ -1201,7 +1201,9 @@ class ADFLOW(AeroSolver):
             # By Default set all the blocks:
             cgnsBlocks = numpy.arange(1, self.adflow.cgnsgrid.cgnsndom + 1)
 
-        self.adflow.preprocessingapi.updaterotationrate(rotCenter, rotations, cgnsBlocks)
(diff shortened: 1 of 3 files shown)


lamkina (changes requested):
A few preliminary things:
1. Please run black on the python layer and fprettify on the fortran layer.
2. For the BC fixes, can you add tests that demonstrate your changes fix the bug?  Testing on a local case won't be enough here because you are changing manually differentated routines.
3. Can you add a test for the rotation rate fix?  I would like to see a test that demonstrates the expected capability and shows your changes achieve that capability.

eirikurj:
I would prefer two separate PRs.

PR #330, opened 2023-11-28
Updates to several overset update routines

anilyil:
## Purpose

This PR contains several changes to the overset cutting algorithm. These can be grouped as:

1. Surface Callback Tracking: In https://github.com/mdolab/adflow/pull/231, I introduced a new way to do the overset hole cutting. This method introduced the use of blanking surfaces that can be used instead of the regular flooding algorithm for difficult cases. In the current PR, I cleaned up that code a bit and enabled "DVGeo tracking" for these surfaces. So as the main CFD mesh is deformed, the blanking surface is deformed consistently. This is useful when we need to re-run the full overset update after geometry changes. The DVGeo embedding is done in an embarrassingly parallel way because this blanking surface is duplicated on all procs.

2. Changes to explicit surface cutting: Again, the feature mentioned above uses an algorithm under the hood to determine which cells are "inside" the closed surface. When these blanking surfaces are thin and they cut large cells relative to their thickness, some cells might slip from the cutting algorithm. To fix this, in this PR, I modified how the cells inside the surface are tagged. I first project the cell centers to the blanking surface. If the projection distance is greater than the max diagonal length of the cell, I end the test for that cell. If the projection distance does not satisfy this condition, we then check if the vector from the cell center to any of its vertices can potentially intersect the surface. If so, this cell is explicitly blanked.
Compared to the old algorithm, this is a more conservative check; however, based on my experience, this approach produces good results and is not overly-conservative. The cells that are tagged should not be included in the solution anyways. 

3. Added 2 new options regarding overset updates: 
 - `updateWallAssociations`: Enables updating the wall associations when geometry is updated, even when we are using the approximate wall distance routines. The overset meshes can only use the approximate wall distance routine, but before this PR, it was hard coded to always keep associations. In this PR, this behavior can be controlled by the user.
- `recomputeoverlapmatrix`: In the full overset update mode, the block overlap matrix was not re-computed. If either the deformations are large, the CGNS block has many blocks, or both, this overlap matrix needs to be re-computed to get a valid hole cutting. In case someone still wants "efficient" full overset updates with small changes, they can use this option to disable the update. The default is set to True because it is somewhat likely for the overlap matrix to change when the mesh has large enough deformations that necessitates a full overset update.

4. Updated how orphan cells are written to the CGNS volume file: Back in 2018, I added the capability to write a -5 for the cells that were considered as orphans in the hole cutting: https://github.com/mdolab/adflow/commit/fe9a0159e8d1d46519b662c2cf79f2ca3baad4b7 However, I don't think this check was correct. There are actually two checks in that routine; the first check is for connectivity problems of compute cells. I think this check is triggered sometimes incorrectly for non-regular CGNS domain topologies (3, 5 point singularities etc). This check is where I was tagging the "orphans". However, I don't think this is the right definition. 
The following check is following the same logic in reverse; it starts from cells that are not compute or interpolate, and goes 2 cells in each direction. If a cell within this "radius" is compute, that cell is called an orphan cell. 
In this PR, I changed the orphan cell definition to be consistent with the second check. That check's result is what is printed as "number of orphans", and my old code would write more "orphan" cells in the volume solution than what is printed, causing confusion.

Unfortunately, I don't have tests for these routines. I am still planning to add some overset tests that go through all of these, but that will require some thinking and time. Until then, it would make my life a lot easier if we merge these changes. 

## Expected time until merged
1-2 months

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


src/overset/oversetAPI.F90
```diff
@@ -2237,51 +2238,82 @@ subroutine flagCellsInSurface(pts, npts, conn, nconn, flag, ncell, blockids, nbl
...
+
+                                ! reset the "closest point already found" variable.
+                                coor(4) = dStar
+                                intInfo(3) = 0
+                                call minDistancetreeSearchSinglePoint(ADT, coor, intInfo, &
+                                                                      uvw, dummy, 0, BB, &
+                                                                      frontLeaves, frontLeavesNew)
+                                cellID = intInfo(3)
+
+                                ! increment the cell counter before the checks in case we exit early
```
sseraj, line 2260: What is this comment referring to?

sseraj (commented):
Great changes! I tested the new explicit hole cutting algorithm on my mesh and can confirm that it is more conservative but also more effective. I just have a couple of comments

adflow/pyADflow.py
```diff
@@ -3345,6 +3298,37 @@ def setAeroProblem(self, aeroProblem, releaseAdjointMemory=True):
...
+                        disp[1:] = numpy.cumsum(sizes)
+
+                        # save this info in the dict
+                        self.blankingSurfData[surfFile]["sizes"] = sizes
+                        self.blankingSurfData[surfFile]["disp"] = disp
+
+                        # we already communicated the points when loading the file,
+                        # so just add them to dvgeo now
+                        procPts = surfPts[disp[self.comm.rank] : disp[self.comm.rank + 1]]
+                        self.DVGeo.addPointSet(procPts, surfPtSetName, **self.pointSetKwargs)
```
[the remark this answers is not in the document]
sseraj: That makes sense, thanks

src/overset/oversetAPI.F90
sseraj: The cell counter logic looks fine

sseraj (dismissed):

src/overset/oversetAPI.F90
```diff
@@ -2237,51 +2238,80 @@ subroutine flagCellsInSurface(pts, npts, conn, nconn, flag, ncell, blockids, nbl
...
+                                cellID = intInfo(3)
+
+                                keepChecking = .True.
+
+                                if (cellID > 0) then
+                                    ! Now check if this was successful or not:
+                                    if (checkInside()) then
+                                        ! Whoohoo! We are inside the region. Flag this cell
+                                        flag(cell_counter) = 1
+                                        keepChecking = .False.
```
eirikurj, line 2267: Is `keepChecking` needed? Seems like these are progressive if checks.

src/overset/oversetUtilities.F90
```diff
@@ -1670,6 +1666,12 @@ subroutine checkOverset(level, sps, totalOrphans, printBadCells)
                                         ! This cell is an orphan:
                                         n = n + 1
                                         orphans(:, n) = (/ii, jj, kk/)
+                                        if (printBadCells) then
```
eirikurj, line 1669: Did you mean to print something here? The logic variables suggest that there is going to be something printed.

eirikurj (changes requested):
Sorry for the delay on this. Please seem my initial comments. At the moment, I dont have an overset mesh to test these changes right now. I think adding tests, even just some simple regression tests that touch this code, if possible, would be great.

eirikurj (changes requested):

src/overset/oversetAPI.F90
```diff
@@ -2237,51 +2238,80 @@ subroutine flagCellsInSurface(pts, npts, conn, nconn, flag, ncell, blockids, nbl
...
+                                coor(4) = dStar
+                                intInfo(3) = 0
+                                call minDistancetreeSearchSinglePoint(ADT, coor, intInfo, &
+                                                                      uvw, dummy, 0, BB, &
+                                                                      frontLeaves, frontLeavesNew)
+                                cellID = intInfo(3)
+
+                                keepChecking = .True.
+
+                                if (cellID > 0) then
```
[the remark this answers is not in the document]
eirikurj: Current implementation is fine for now, but what I initial had in mind is instead of having these large if blocks with lots of code, it would also be possible to do something like the snippet below. The earlier if check `if (k_cgns .gt. k_min) then` could also be wrapped like this. Might make the code a bit more readable.
```
if (cellID <= 0) then
   cell_counter = cell_counter + 1 
   cycle
fi
```

src/overset/oversetUtilities.F90
eirikurj: Ok, I see yeah, I think breaking this up into smaller functions with clear responsibilities, i.e., as separate call might make sense, but we can keep for now.

eirikurj (approved):

sseraj:
Yes, I'll take another look at this and try to run my case as well

sseraj (approved):
The changes look good, and my case still runs the same as when I approved this PR previously

sseraj:
Feel free to merge this (I would, but I took myself out of the maintainers group)

PR #343, opened 2024-02-15
Write failed solution when `restart_failed_analysis` is False

eytanadler:
## Purpose

In the MPhys wrapper, there is an option to try clean restarting an analysis that stalls/fails to converge. In those cases, the failed solution is written so the user can try to figure out what went wrong. However, no failed solution is written when `restart_failed_analysis` is False (this is the default), making diagnosis challenging. This PR changes it so a solution is written in this case.

## Expected time until merged

A few days

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Run an MPhys ADflow case where the `restart_failed_analysis` option in the `ADflowBuilder` is set to `False` and set the max number of ADflow iterations to a very small number so it fails to converge. Before this PR, you will see no failed analysis solution written. With the PR, you will.

## Checklist

### DIFF
--- a/adflow/mphys/mphys_adflow.py
+++ b/adflow/mphys/mphys_adflow.py
@@ -555,6 +555,11 @@ class ADflowSolver(ImplicitComponent):
                         print("###############################################################")
                         print("# Solve Failed, not attempting a clean restart")
                         print("###############################################################")
+
+                    # write the solution so that we can diagnose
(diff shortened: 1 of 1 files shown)


lamkina (approved):
Typically we just catch the error in our scripts and then write a solution, but I think this is fine because it's not incurring any massive memory overhead to write a single failed solution.

PR #354, opened 2024-05-14
Multiple BC Fix

lamkina:
## Purpose

This PR addresses a bug where adding multiple BCs of the same type didn't work as expected. Previously, only the first BC of a given type would be properly added and tracked, but then any subsequent BC of the same type would be ignored. This resulted in incorrect functionals and derivatives with zero seeds.

Using the suggested fixes from PR #323 this should now be fixed. I added a new mesh to `input_files` and tests to make sure this is working correctly.

## Expected time until merged

1-3 weeks

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Run `testflo tests/reg_tests/test_multiple_bc.py` to test the changes. You can also run the complex tests in the same file to verify the derivatives are working as expected.

## Checklist


lamkina:
@sseraj @eirikurj The failing tests are passing locally so I'm not sure what is happening. This specific test is just checking the totals of BC integrated functions computed with the adjoint against the totals that are saved in the training data. Locally it matches to 1e-12 absolute and relative tolerance, but on the images it's at worst getting 1e-6 absolute and 1e-5 relative.

Any debugging tips on what might be happening?

sseraj:
I have had tests fail like this when the flow converges fine locally but does not fully converge on the Docker images. I would check that the flow and adjoint both converge to the same level on the Docker images as on your local machine. You might have to increase `nCycles`

lamkina:
Following the discussion in #358 we should update the patch version in this PR when it's eventually ready.

Issue #324, opened 2023-10-11
AMG regression test fails with three levels


eirikurj:
## Description
Nightly docker integration tests on the `u22-gcc-ompi-stable` x86 image is failing on the AMG test (see [here](https://github.com/mdolab/docker/actions/runs/6478172332/job/17603179934#step:13:1198)). This seems to be a result of a newly merged PR https://github.com/mdolab/adflow/pull/235. 

The `test_amg.py:TestAMGReal_2_different_levels_and_iterations.test_convergence` AMG regression test is failing to converge.

### Steps to reproduce issue
This can be reproduced with the failed image pushed to DockerHub, i.e., `mdolab/public:u22-gcc-ompi-stable-amd64-failed` (`Digest: sha256:76e6eb4be1c823744ec7cfc46fa38e5edc9897b8cf9d54dc08d208cd5db1d2f2`). Using the image, download the input test files, and run only the AMG regression tests `testflo -s -v -n 1 test_amg.py`.

Using the previous commit `git checkout 559a1deee07126c521e7a2e18f3968dd0ef7b061`, rebuilding and pip installing, then all AMG tests do pass successfully.

### Current behavior
Currently, the case does not converge and test fails.

### Expected behavior
Case should converge and test pass. 

### Code versions
Docker image `mdolab/public:u22-gcc-ompi-stable-amd64-failed` downloaded today (`Digest: sha256:76e6eb4be1c823744ec7cfc46fa38e5edc9897b8cf9d54dc08d208cd5db1d2f2`)

sseraj:
This was fixed in #325 by changing the test. The root cause is unclear except that using too many AMG levels potentially causes memory-related issues.

eirikurj:
@sseraj, if there is a known issue, I would prefer to either document it somewhere in the docs or create a separate issue.

sseraj:
This issue has all the relevant information, so maybe it is best to just reopen this

PR #359, opened 2024-05-31
Coarse level options for AMG preconditioner

sseraj:
## Purpose

This PR adds separate ASM overlap, ILU fill level, and inner (ILU) iteration options for the coarse levels when using the AMG preconditioner. The motivation behind this is to allow for cheaper iterations at the coarse levels to speed up the overall solution time. The defaults for the coarse levels are set to the cheapest options possible (no ASM overlap, no ILU fill, 1 inner iteration). This changes the default behavior but should be advantageous for most cases.

## Expected time until merged

2-3 weeks

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added an AMG test with non-default options. The default options are used in the existing tests.

## Checklist


sseraj:
I think it would be better to make the ANK and NK changes in a separate PR

lamkina (dismissed):

lamkina:
@eirikurj I think this is good to go, but we can wait to merge if you want to take a look.

tests/reg_tests/test_amg.py
```diff
@@ -82,6 +82,27 @@
             "innerPreconIts": 3,
         },
     },
+    {
```
eirikurj, line 85: Seems like we are only checking convergence for these tests. These values here may not reflect best practice, but do we not want to track some reference values, either convergence rate (or some functional output) as well (or both)? If not now, we should add such tests in the future (create issue with improvements).

eirikurj (changes requested):

doc/options.yaml
```diff
@@ -1464,18 +1488,30 @@ ILUFill:
...
+
 innerPreconIts:
   desc: >
     Number of local preconditioning iterations for the adjoint solution.
     Increasing this number may help with difficult problems.
     However, each iteration will take more time.
 
+innerPreconItsCoarse:
+  desc: >
+    Same as :py:data:`innerPreconItsCoarse` but for the coarse levels when using ``multigrid`` for :py:data:`globalPreconditioner`.
```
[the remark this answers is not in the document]
sseraj: Fixed, thanks

tests/reg_tests/test_amg.py
sseraj: I'm not sure if it's worth checking functional outputs because that doesn't really test the solver. For convergence rate, we could track the number of iterations or time, but both of those will likely depend on the image the test is running on.

eirikurj: Right, I understand, I will approve this for now. However, I do think we need to improve the tests sooner than later (especially we have an open [issue](https://github.com/mdolab/adflow/issues/324) on the AMG). 
@sseraj can you maybe create an issue, suggesting updates that we can make to improve these solver tests?

eirikurj (approved):

lamkina (approved):

tests/reg_tests/test_amg.py
sseraj: I opened an issue here: #361

PR #362, opened 2024-08-30
Added new tests for two consecutive adjoint runs

sabakhshi:
## Purpose
In the development effort for SST, it was discovered that Tapenade was causing changes to the solutions states after an adjoint run due to a bug. To ensure that this was not a systematic problem throughout the code and to catch future instances of this issues appearing, I added two new adjoint tests. The first runs the adjoint twice back to back to ensure that the results didn't change. The second runs the adjoint twice back to back and checks to make sure that the states have not changed between the runs.

## Expected time until merged
2-3 weeks

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
The new feature is a new test.

## Checklist


tests/reg_tests/reg_test_utils.py
```diff
@@ -13,6 +13,24 @@ def assert_adjoint_sens_allclose(handler, CFDSolver, ap, evalFuncs=None, **kwarg
     handler.root_print("Eval Functions Sens:")
     handler.root_add_dict("Eval Functions Sens:", funcsSens, rtol=rtol, atol=atol)
 
+def assert_adjoint2_sens_allclose(handler, CFDSolver, ap, evalFuncs=None, **kwargs):
+    rtol, atol = getTol(**kwargs)
+    funcsSens = {}
+    CFDSolver.evalFunctionsSens(ap, funcsSens, evalFuncs=None)
+    CFDSolver.evalFunctionsSens(ap, funcsSens, evalFuncs=None)
+    handler.root_print("Eval Functions Sens 2:")
+    handler.root_add_dict("Eval Functions Sens 2:", funcsSens, rtol=rtol, atol=atol)
```
marcomangano, line 22: Shouldn't we compare the second run against the values of the first run `Eval Functions Sens`? I don't see the point of adding this if the point of the test is to make sure that the adjoint converges to the same result (within a tolerance)

tests/reg_tests/reg_test_utils.py
```diff
@@ -43,7 +61,7 @@ def assert_residuals_allclose(handler, CFDSolver, ap, **kwargs):
     res /= totalR0
     handler.root_print("Norm of residual")
     handler.par_add_norm("Norm of residual", res, rtol=rtol, atol=atol)
-
+
```
marcomangano, line 64: This is probably causing the black and flake8 error

tests/reg_tests/refs/adjoint_euler_scalar_jst_tut_wing.json
```diff
@@ -1,86 +1,87 @@
 {
-    "Eval Functions Sens:": {
+    "Adjoint States": 0.0,
+    "Eval Functions Sens 2:": {
```
marcomangano, line 3: As I mention below, I don't think we should add this, ideally the only thing added to the `.json` files is the `Adjoint States` entry

marcomangano (changes requested):
See my comments below, I don't think we should add separate reference results for the second sensitivity analysis, am I missing anything?

tests/reg_tests/refs/adjoint_euler_scalar_jst_tut_wing.json
```diff
@@ -1,86 +1,87 @@
 {
+    "Adjoint States": 0.0,
```
marcomangano, line 2: This entry should be the only one changing in the `.json` files. There is no point changing the reference values for this PR, especially if the training is done on a local machine.

marcomangano (changes requested):
The added tests look good, I only think the `.json` files need to be cleaned up - see my comment below

marcomangano:
Also note that the black and flake8 tests are failing, please fix them. Remember that we currently use `black==23.1.0` and custom `flake8` settings

sabakhshi:
Everything should be addressed. Re-requesting your review.

marcomangano (approved):
Nice cleanup, thanks!

lamkina (approved):

PR #375, opened 2025-01-17
Add improved actuator zone model

anilyil:
## Purpose

This adds the improved actuator zone model that includes more parameters to control the distribution of the actuator zone terms, as well as adding a swirl term. I believe some of the changes originated from the work of @shamsheersc19.

It is WIP right now but opening the PR to start the discussions on this. 

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/adflow/actuatorRegion.py
+++ b/adflow/actuatorRegion.py
@@ -169,7 +169,7 @@ class ActuatorRegionHandler:
                 heat.T,
             )
 
-    def writeActuatorRegions(self, adflow_fortran: ModuleType, fileName: str, outputDir: str):
+    def writeActuatorRegions(self, fileName: str, outputDir: str):
         # Join to get the actual filename root
         fileName = os.path.join(outputDir, fileName)
 
@@ -178,7 +178,7 @@ class ActuatorRegionHandler:
         fileName += ".plt"
 
         # just call the underlying fortran routine
-        adflow_fortran.actuatorregion.writeactuatorregions(fileName)
+        self._adflow_fortran.actuatorregion.writeactuatorregions(fileName)
 
 
 class CircularActuatorRegion(AbstractActuatorRegion):


eytanadler:
Currently, the actuator zone's thrust cannot be set through the `AeroProblem` as expected if the actuator zone uses the `simpleProp` type. The `simpleProp` uses the `thrust` attribute from `actuatorRegionData` rather than the `force` one in the residual routine. However, only the `force` attribute is set [in the update BC call](https://github.com/anilyil/adflow/blob/ef0a338e8adfd5332a020836216c8f71c6934ad3/src/bcdata/BCData.F90#L1488). To fix this, the `thrust` attribute must also be updated with something like `actuatorRegions(iRegion)%thrust = bcDataIn(iVar)`. This could be done in an if/else based on the actuator region type. I'd make this change, but I don't have Tapenade set up to differentiate it. @eirikurj do you have an environment set up where you could easily make this change?

eirikurj:
@eytanadler have everything setup and can add this, but I dont have any case available to test this. Do you have anything that can be considered as a unittest?

eytanadler:
I don't have anything unit testable but the basic behavior necessary is that this wasn't working before:
```python
solver = ADFLOW(options=options)
solver.addActuatorRegion(
    "az_surf.x",
    actType="simpleProp",
    axis1=np.array([0.0, 0.0, 0.0]),
    axis2=np.array([1.0, 0.0, 0.0]),
   familyName="az",
    thrust=0.0,
    propRadius=1.0,
    spinnerRadius=0.1,
)

# Doing this should make thrust settable through the AeroProblem
ap.setBCVar("Thrust", 0.0, "az")
ap.addDV("Thrust", family="az", name="az_thrust")
ap.setDesignVars({"az_thrust": 1e4})

solver(ap)
```
Specifically, adding the design variable to the `AeroProblem` and setting it should change the thrust (observable via `flowpower` cost function), but it doesn't without the changes I mentioned. I think something like this would make a fine unit test, right?

eytanadler:
I just noticed that I don't think the `mDistribParam` and `mDistribParam` match what is in [Sham's paper](https://www.researchgate.net/profile/Shamsheer-Chauhan/publication/350073795_RANS-Based_Aerodynamic_Shape_Optimization_of_a_Wing_Considering_Propeller-Wing_Interaction/links/61630b65ae47db4e57bc9760/RANS-Based-Aerodynamic-Shape-Optimization-of-a-Wing-Considering-Propeller-Wing-Interaction.pdf). His paper has essentially $f_x = r^m (1 - r)^n$ (I'm simplifying here), but [the code](https://github.com/anilyil/adflow/blob/4745f0cea2183c56496143c435abe4f0ee6bbace/src/solver/actuatorRegion.F90#L279) looks like it implements something more like $f_x = r^m (1 - r)^n / r$ with the extra `region%cellRadii(region%nCellIDs)` in the denominator. This reduces $m$ by 1 relative to the equation in the paper. This isn't necessarily a problem, but it's not intuitive. I would suggest either making the small tweak to the implementation to match the paper or changing the default `mDistribParam` to be 2.0 instead of 1.0 and putting a comment in the docstring because the default distribution right now not very reasonable. Am I missing something here?

shamsheersc19:
I'm not sure if you deleted the comment about the differences in the
formulas in the code vs the paper, Eytan. I cannot find it on GitHub, but I
see it here in my email. In *fact = one / propRadius*, I think I had the
extra radius in the denominator to make the very small numbers larger for
the summing and dividing later. In the end, the extra divide by propRadius
(which is the outer radius i.e., a constant value) gets cancelled out in
the step later where *region%thrustVec = region%thrustVec /
region%totalThrustSum*. The *(two * pi * region%cellRadii(region%nCellIDs))
*is to convert the force per unit radius (which is what the formula gives)
to per unit volume, as described in the paper. I haven't included the disk
thickness in the denominator in the code because it is a constant that
cancels out in the later step I mentioned earlier.

I need to think more about what consequences that dot product error that
you fixed will have.

shamsheersc19:
I had a look at the dot product error fixed in commit [0d21897](https://github.com/mdolab/adflow/pull/375/commits/0d2189737d726278c74748f59e5eb23eea04fe19). I got lucky there because in the cases I ran for the papers, v1(1) = 0 (or approximately 0) so it being multiplied by the incorrect number fortunately didn't affect my cases. Nice catch!

PR #396, opened 2026-02-27
fix TS adjoint

SichengHe:
## Purpose

Fix the time spectral (TS) adjoint, which was completely broken because the `case (timeSpectral)` block in `initres_block` (`src/solver/residuals.F90`) was wrapped inside `#ifndef USE_TAPENADE`, so Tapenade never generated adjoint code for the spectral time derivative terms. The generated AD code only handled `case (steady)`, completely missing the TS contributions.

Changes:
- Exposed `case (timeSpectral)` in `residuals.F90` to Tapenade by moving the `#endif` boundary, with `#ifdef USE_TAPENADE` blocks to replace pointer access with direct `flowDoms` array access (required since Tapenade cannot differentiate through Fortran pointers)
- Added reverse mode adjoint for TS in `residuals_b.f90`
- Added forward mode tangent for TS in `residuals_d.f90`
- Added state-only reverse mode for TS in `residuals_fast_b.f90`
- Added test script `tests/test_ts_adjoint.py` for verification

## Expected time until merged

Non-urgent. A week or two is fine.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Run with 2 MPI ranks on the NACA 64A010 Euler time spectral case (3 time intervals, pitching oscillation):
mpirun -np 2 python tests/test_ts_adjoint.py

Three tests are performed:
1. **Dot product test** (forward vs backward consistency): relative error = 1.06e-15 (machine precision)
2. **Forward AD vs finite difference**: residual derivative relative error = 4.03e-05; function derivatives (cl, cd, cmz) match to 1e-7 — 1e-8
3. **Adjoint sensitivity** (`evalFunctionsSens`): all three adjoint solves (cd, cl, cmz) converge

## Checklist

- [ ] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted


tests/test_ts_adjoint.py
A-CGray: Please put this file in the `reg_tests` directory

A-CGray (changes requested):
Thanks for these fixes @SichengHe , can you please update your tests to use ADflow's existing `RegTest` class. See [this file](https://github.com/mdolab/adflow/blob/main/tests/reg_tests/test_cavitation.py) for an example.

Also, which version of Tapenade did you use to generate the new derivative code? I think the failing Tapenade check is probably due to a version mismatch, the test uses 3.16

sanjan98 (dismissed):
Time spectral adjoint fixed.
Moves `#endif` before `case(timeSpectral)`
`#ifdef` guards

tests/reg_tests/refs/euler_time_spectral_naca64A010.json
```diff
@@ -1,32 +1,47 @@
 {
+    "Adjoint States": 0.0,
+    "Eval Functions Sens:": {
+        "64A010pitchingTS_cd": {},
+        "64A010pitchingTS_cl": {},
+        "64A010pitchingTS_cmz": {}
+    },
```
A-CGray, line 7: No derivatives are being computed because you don't have any DVs defined

A-CGray (changes requested):
I combined the test file you added with the existing time spectral test because all of the setup code was the same.

In the PR summary you say:

> Three tests are performed:
> 
> - Dot product test (forward vs backward consistency): relative error = 1.06e-15 (machine precision)
> - Forward AD vs finite difference: residual derivative relative error = 4.03e-05; function derivatives (cl, cd, cmz) match to 1e-7 — 1e-8
> 

However, I don't see anything in the new time spectral tests that check these

A-CGray:
@SichengHe , reminder about this

JavierAlbaM:
@SichengHe Hey! If you have the chance, please address the comments within the next 2 weeks, otherwise we will close the PR or send it to draft. Thanks!

PR #401, opened 2026-04-10
Fixed an MPhys issue with setting scalars with Numpy 2.4

sabakhshi:
## Purpose
The Mphys wrapper was not working with Numpy versions >2.4.
This is because as of Numpy 2.4 1D vectors of length 1 can no longer be treated as scalar data types.
This issue manifested when ADflow tried to print the alpha value set by OpenMDAO.
We now check if we are setting a 1D array of len(1) and explicitly index the first entry to get a scalar 
data type.

## Expected time until merged
1-2 days

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
No new test required since we don't test on Numpy 2.4 yet.
This is just getting ahead of the issue.
The fix may no longer be needed if OpenMDAO changes something upstream
to account for this.

## Checklist

- [x] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted

### DIFF
--- a/adflow/mphys/mphys_adflow.py
+++ b/adflow/mphys/mphys_adflow.py
@@ -206,7 +206,11 @@ def setAeroProblem(solver, ap, ap_vars, inputs=None, outputs=None, print_dict=Tr
         tmp = {}
         for args, _ in ap_vars:
             name = args[0]
-            tmp[name] = inputs[name]
+            # This is a fix to support Numpy 2.4 until OpenMDAO supports it
+            if np.shape(inputs[name]) == (1,):
+                tmp[name] = inputs[name][0]
+            else:
+                tmp[name] = inputs[name]
 
         ap.setDesignVars(tmp)
         if solver.comm.rank == 0 and print_dict:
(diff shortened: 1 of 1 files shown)


A-CGray (approved):
Thanks for fixing this @sabakhshi

A-CGray:
@sanjan98 you're running ADflow through MPhys at the moment right? Could you test out this fix on your case using numpy 2.3 and 2.4?

sanjan98:
@A-CGray I can test it tomorrow and approve

sanjan98:
@sabakhshi @A-CGray 
Going through this PR, isn't it better to fix the root cause at the OpenMDAO level? Or add numpy2.4 to the CI testing matrix if we dont want to depend on OpenMDAO? I dont understand how I can isolate testing for just this fix with numpy 2.4. Let me know what you guys think.

sabakhshi:
Fixing the root cause at the OpenMDAO level will take a while since this type of casting that was dropped in Numpy 2.4 is heavily used there. This is more of a band aid fix until the OM team can get around to it. We don't need to officially test in the CI matrix with Numpy 2.4 since we don't officially claim support for Numpy 2.4. This is just more a solution to allow people get by with  Numpy 2.4 until official support comes in from OM.

sanjan98 (approved):
Approving changes after discussion.

PR #405, opened 2026-06-22
Fix potential stack overflow in overset initialisation

A-CGray:
## Purpose

I have ADflow compiled locally with `-Ofast` optimisations and was running into segfaults when trying to run certain overset meshes. Turns out `-Ofast` enables the `-fstack-arrays` [flag](https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html#index-fstack-arrays), which causes gfortran to allocate the memory for arrays on the stack, even when it doesn't know their size. Depending on the size on the size of your mesh, and the number of cores you're working with, some of the code in the overset utils ends up allocating arrays on the stack which are too large (the default limit is 8mb apparently), leading to a stack overflow.

This PR fixes the issue by switching one of the problematic array copies to an explicit do loop, and replaces the other ones with array reallocation utility routines that are already in the code (that also use an explicit do loop).

This shouldn't be a commonly encountered issue because it only happens with `-Ofast` and relatively high (~150k) cell counts per core.

## Expected time until merged

Soon

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted

### DIFF
--- a/src/overset/oversetUtilities.F90
+++ b/src/overset/oversetUtilities.F90
@@ -1240,7 +1240,7 @@ contains
         integer(kind=intType), intent(inout) :: n
 
         ! Working Paramters
-        integer(kind=intType) :: fSize
+        integer(kind=intType) :: fSize, i
         type(fringeType), dimension(:), pointer :: tmpFringePtr
         fSize = size(fringeList)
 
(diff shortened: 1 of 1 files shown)


sanjan98:
@A-CGray do we need a separate unit test for this compilation check? I am not sure how to test for a compilation error specific to a certain flag.

sanjan98 (approved):
Haven't tested it as it is a compilation check but Ali says he checked it

